PR #2, opened 2021-05-26
Push initial csi-lib code/implementation

munnerz:
This is a push of the initial implementation of a standalone CSI library that can be used as a basis for new CSI drivers that interact with cert-manager APIs to fetch identity documents.

There are a number of follow up PRs that we have in-flight that I'll also push up once this has merged.

cc @JoshVanL


JoshVanL:
/lgtm
/approve
/hold

Unhold when you're ready @munnerz

PR #3, opened 2021-05-26
Support deleting/cleaning up CertificateRequest objects for volumes

munnerz:
This implements the cleanup portion of what's described in the `certificaterequest-management.md` design document.

It allows the driver to delete 'old' CertificateRequest resources (up to a configurable limit) to avoid creating an excessive number of resources, impacting the stability of the apiserver.

### DIFF
--- a/internal/api/consts.go
+++ b/internal/api/consts.go
@@ -1,0 +1,6 @@
+package api
+
+const (
+	NodeIDHashLabelKey   = "csi.cert-manager.io/node-id-hash"
+	VolumeIDHashLabelKey = "csi.cert-manager.io/volume-id-hash"
+)
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -99,11 +99,15 @@ func NewManager(opts Options) (*Manager, error) {
 	if len(opts.NodeID) == 0 {
 		return nil, errors.New("NodeID must be set")
 	}
+	nodeNameHash := hashString(opts.NodeID)
+	nodeNameReq, err := labels.NewRequirement("csi.cert-manager.io/node-name-hash", selection.Equals, []string{nodeNameHash})
+	if err != nil {
+		return nil, fmt.Errorf("building node name label selector: %w", err)
+	}
 
 	// Create an informer factory
 	informerFactory := cminformers.NewSharedInformerFactoryWithOptions(opts.Client, 0, cminformers.WithTweakListOptions(func(opts *metav1.ListOptions) {
-		// TODO: filter informer
-		opts.LabelSelector = ""
+		opts.LabelSelector = labels.NewSelector().Add(*nodeNameReq).String()
 	}))
 	// Fetch the lister before calling Start() to ensure this informer is
 	// registered with the factory
@@ -129,7 +133,7 @@ func NewManager(opts Options) (*Manager, error) {
 		stopInformer:   stopCh,
 
 		maxRequestsPerVolume: opts.MaxRequestsPerVolume,
-		nodeNameHash:         hashString(opts.NodeID),
+		nodeNameHash:         nodeNameHash,
 	}
 
 	vols, err := opts.MetadataReader.ListVolumes()
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -5,7 +5,6 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"hash/fnv"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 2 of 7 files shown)


jtweaver (approved):
/lgtm
/hold

manager/manager.go
```diff
@@ -75,11 +90,24 @@ func NewManager(opts Options) (*Manager, error) {
 	if opts.WriteKeypair == nil {
 		return nil, errors.New("WriteKeypair must be set")
 	}
+	if opts.MaxRequestsPerVolume == 0 {
+		opts.MaxRequestsPerVolume = 3
+	}
+	if opts.MaxRequestsPerVolume < 0 {
+		return nil, errors.New("MaxRequestsPerVolume cannot be less than zero")
+	}
```
munnerz, line 98: We may want to consider permitting e.g. `-1` as a value here to disable cleanup altogether. I've opted to omit that option for now though until we actually have someone request it/have a use-case 😄 (e.g. external cleanup controller)

internal/api/consts.go
```diff
@@ -0,0 +1,6 @@
+package api
+
+const (
+	NodeIDHashLabelKey   = "csi.cert-manager.io/node-id-hash"
+	VolumeIDHashLabelKey = "csi.cert-manager.io/volume-id-hash"
```
munnerz, line 5: VolumeID is expected to base unique *within* a node - so we have both the volume ID and the node ID as part of the 'key' when looking up what is a candidate to delete

manager/manager.go
```diff
@@ -261,6 +303,58 @@ func (m *Manager) issue(volumeID string) error {
 	return nil
 }
 
+func (m *Manager) cleanupStaleRequests(ctx context.Context, log logr.Logger, volumeID string) error {
+	sel, err := m.labelSelectorForVolume(volumeID)
+	if err != nil {
+		return fmt.Errorf("internal error building label selector - this is a bug, please open an issue: %w", err)
+	}
+
+	reqs, err := m.lister.List(sel)
```
munnerz, line 312: Quite note: we do a query across _all_ namespaces here, because there's no way for us to know what namespace resources are being created in until `generateRequest` is called (and some weird implementations could do something wacky like create it in a different namespace each time).

I don't think this is much of an issue though, given we match on the `volumeID` and `nodeID` meaning there can't really be a conflict/problem with this approach.

The only oddity I _can_ think of, is if someone were to compromise a single node, they could craft requests to the kubelet to trigger the driver to delete CSRs for some other volume (though it'd still have to be on the same node, as the node ID is also part of the hash).

So all in all, I don't think this represents any kind of additional attack vector 😄

manager/manager.go
```diff
@@ -75,11 +90,24 @@ func NewManager(opts Options) (*Manager, error) {
 	if opts.WriteKeypair == nil {
 		return nil, errors.New("WriteKeypair must be set")
 	}
+	if opts.MaxRequestsPerVolume == 0 {
+		opts.MaxRequestsPerVolume = 3
+	}
```
JoshVanL, line 95: This is surprising to me. What if I did want this to be 0. Does it make sense for 0 to be an option?

I understand that we can't make use of defaulting in the same way a CLI flag can, and making the type a pointer seems clumsy.

munnerz: 0 would imply never leaving a CertificateRequest there at all, e.g. it is deleted immediately after creation/use, which I don't think makes sense (and would take some tweaking of the way cleanup is invoked to make possible).

Perhaps this _is_ something we want to explore in future, but I'd imagine really people want `1` (i.e. only ever have 1 CR for this volume at any given time).

I'm definitely open to expanding the allowed options here if we have a use-case. We should decide whether `0` does make sense as an option, because changing this in future could be painful (as users who don't configure this field explicitly will have a max of 3 enforced, so we won't easily be able to change this to have different behaviour in future without making a breaking Golang API change)

munnerz: If a CertificateRequest *fails* and this is set to zero, would we expect it to be immediately deleted too?

manager/manager.go
```diff
@@ -75,11 +90,24 @@ func NewManager(opts Options) (*Manager, error) {
 	if opts.WriteKeypair == nil {
 		return nil, errors.New("WriteKeypair must be set")
 	}
+	if opts.MaxRequestsPerVolume == 0 {
+		opts.MaxRequestsPerVolume = 1
```
JoshVanL, line 94: After some discussion, defaulting to 1 makes sense here. Anything above 1 seems to me like a special case, or temporarily set for debugging purposes. I would expect this value to be typically bubbled up to an end user anyway, where an author can enforce their own defaulting using a CLI flag. An author not exposing this option to the user will likely  be opinionated about the value anyway and set it themselves.

We can revisit this later to decide if and how we can represent a "clean up after use", for the cases where the driver should delete the CR immediately after it has been signed and consumed. There is also the option to _never_ cleanup CRs. In these cases we can either decide to utilise special negative numbers like below, and/or make a breaking API change to change this field type to be a pointer.

```go
const MaxRequestPerVolumeCleanupAfterUse = -1
```

PR #6, opened 2021-05-26
Expect PEM data to be returned by SignRequestFunc instead of DER

munnerz:
This PR changes the public/exported interface (that consumers will implement) for the library to shift the responsibility for encoding to PEM to the implementor of the library.

This is to allow implementors to add e.g. trailing PEM comments (which are permitted by both the PEM spec and the Certificate(Signing)Request API). These may be used for any purpose, including (but not limited to) attaching additional attestation information that does not fit within the x509 document.

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -191,7 +191,15 @@ func generateRequest(meta metadata.Metadata) (*manager.CertificateRequestBundle,
 }
 
 func signRequest(_ metadata.Metadata, key crypto.PrivateKey, request *x509.CertificateRequest) ([]byte, error) {
-	return x509.CreateCertificateRequest(rand.Reader, request, key)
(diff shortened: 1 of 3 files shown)


JoshVanL:
Looks good to me, this is what I would expect as a developer.

/lgtm

munnerz:
@jtweaver this is a breaking change to our Golang API and will require any implementations of the driver to update their code to encode the CSR bytes as PEM before returning (in `SignRequest`) FYI

munnerz:
Wake up Tide, please? 🙄 🤖

SgtCoDFish:
here goes nothing:

/hold

JoshVanL (approved):

PR #7, opened 2021-05-26
Add exponential back-off when issuance fails

munnerz:
This patch adds exponential back-off when issuance is required for a certificate.

This should not be merged yet as it needs more extensive testing (both automated and manual). I'm pushing it up to get some additional eyes on it, and in case anyone wants to test it out. I'm not too familiar with the exponential back-off functions in the `wait` package, and the numbers I have chosen to configure the back-off are a little hand wavy 😄

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"sort"
 	"sync"
 	"time"
@@ -490,7 +491,11 @@ func (m *Manager) ManageVolume(volumeID string) error {
 						Duration: time.Second * 2,
 						Factor:   2.0,
 						Jitter:   0.5,
-						Steps:    6,
+						// Set this to the maximum int value to avoid resetting the exponential backoff
+						// timer early.
+						// This will mean that once the back-off hits 1 minute, we will constantly retry once
+						// per minute rather than resetting back to `Duration` (2s).
+						Steps:    math.MaxInt32,
 						Cap:      time.Minute,
 					}, func() (bool, error) {
 						log.Info("Triggering new issuance")
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -460,6 +460,15 @@ func (m *Manager) ManageVolume(volumeID string) error {
 	stopCh := make(chan struct{})
 	m.managedVolumes[volumeID] = stopCh
 
+	// Create a context that will be cancelled when the stopCh is closed
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		select {
+		case <-stopCh:
+			cancel()
+		}
+	}()
+
 	go func() {
 		// check every volume once per second
 		// TODO: optimise this to not check so often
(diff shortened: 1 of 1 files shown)


manager/manager.go
```diff
@@ -349,13 +358,20 @@ func (m *Manager) ManageVolume(volumeID string) error {
...
-						// retry the request in 1 second time
-						// TODO: exponentially back-off
-						continue
-					}
+					wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
+						Duration: time.Second * 2,
+						Factor:   2.0,
+						Jitter:   0.5,
+						Steps:    6,
+						Cap:      time.Minute,
```
munnerz, line 366: This says:

* 2s is the 'base' amount of time to back off
* multiply the previous value by 2 if it fails
* add a jitter of +/- 1s (half of the `duration`)
* 6 steps (this actually may need tweaking.. it may mean that it resets back to 0 after 6 'backoffs' - need to test)
* with a max time between calls of 1m

maelvls: suggestion: add your very good explanation as a comment on top of `wait.ExponentialBackoffWithContext` 👍👍

manager/manager.go
```diff
@@ -332,6 +332,15 @@ func (m *Manager) ManageVolume(volumeID string) error {
...
 	m.managedVolumes[volumeID] = stopCh
 
+	// Create a context that will be cancelled when the stopCh is closed
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		select {
+		case <-stopCh:
+			cancel()
+		}
+	}()
```
maelvls, line 342: question: could we use a context instead of `stopCh`?

```go
	ctx, cancel := context.WithCancel(context.Background())
	m.managedVolumes[volumeID] = ctx.Done
```

maelvls: Never mind, I just read that `ctx.Done` can't be closed since it is "managed" by the Context package 😇

manager/manager.go
```diff
@@ -477,13 +487,24 @@ func (m *Manager) ManageVolume(volumeID string) error {
...
 
 				if meta.NextIssuanceTime == nil || m.clock.Now().After(*meta.NextIssuanceTime) {
-					log.Info("Triggering new issuance")
-					if err := m.issue(volumeID); err != nil {
-						log.Error(err, "Failed to issue certificate")
-						// retry the request in 1 second time
-						// TODO: exponentially back-off
-						continue
-					}
+					wait.ExponentialBackoffWithContext(ctx, wait.Backoff{
```
JoshVanL, line 490: Change seems good to me and happy to merge.

The second here is not great. What is stopping us constructing a timer at every loop, which takes `*meta.NextIssuanceTime` as input? Can this value change outside of `m.issue`?

JoshVanL:
@munnerz comment from me, feel free to unhold

/lgtm
/hold

manager/manager.go
munnerz: Agreed - I'll be following up with more PRs to adjust this timer to not just be a fixed 1s, as we can be more intelligent than that :)

manager/manager.go
munnerz: I've just unheld this for merge, but +1 and I'll open a PR in a sec with this :)

PR #9, opened 2021-05-31
Adds ClientForMatadataFunc to manager 

JoshVanL:
This PR adds an optional option to the manager: `ClientForMetadata`. This takes a function which returns a cert-manager client based on the passed Volume metadata. This is designed to be used to generate on the fly cert-manager clients that make use of [CSI token requests](https://kubernetes-csi.github.io/docs/token-requests.html). CertificateRequests created by this client will then have the UserInfo of the pod who owns that Volume. This allows a CSI driver to request identity documents on the pods behalf, whilst an approver can validate the contents of the certificate request matches that of the pod.

By default this is unset, which means the `Client` cert-manager client will be used to create CertificateRequests.

I have modified the filesystems `RegisterMetadata` func so that it will write to file also when the volume context has changed. This is important as the Kubelet will call `NodePublishVolume` again when the current token has expired, passing a new one. Without this change, the driver would continue to use a stale token which would result in certificate renewals being rejected.

---

Sets `success` to `true` when `targetPath` is already a bind point to prevent Unmounting an already successful mount point.

/cc @munnerz

### DIFF
--- a/manager/interfaces.go
+++ b/manager/interfaces.go
@@ -23,6 +23,7 @@ import (
 
 	cmapi "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
 	cmmeta "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
+	cmclient "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
 
 	"github.com/cert-manager/csi-lib/metadata"
 )
@@ -82,3 +83,8 @@ type SignRequestFunc func(meta metadata.Metadata, key crypto.PrivateKey, request
 // The 'chain' and 'ca' arguments are PEM encoded and sourced directly from the
 // CertificateRequest, without any attempt to parse or decode the bytes.
 type WriteKeypairFunc func(meta metadata.Metadata, key crypto.PrivateKey, chain []byte, ca []byte) error
+
+// CreateClientFunc will return a cert-manager API client used for creating
+// objects. This is called with the metadata associated with the volume being
+// published. Useful for modifying clients to make use of CSI token requests.
+type CreateClientFunc func(meta metadata.Metadata) (cmclient.Interface, error)
--- a/manager/interfaces.go
+++ b/manager/interfaces.go
@@ -84,7 +84,8 @@ type SignRequestFunc func(meta metadata.Metadata, key crypto.PrivateKey, request
 // CertificateRequest, without any attempt to parse or decode the bytes.
 type WriteKeypairFunc func(meta metadata.Metadata, key crypto.PrivateKey, chain []byte, ca []byte) error
 
-// CreateClientFunc will return a cert-manager API client used for creating
-// objects. This is called with the metadata associated with the volume being
-// published. Useful for modifying clients to make use of CSI token requests.
-type CreateClientFunc func(meta metadata.Metadata) (cmclient.Interface, error)
+// ClientForMatadataFunc will return a cert-manager API client used for
+// creating objects. This is called with the metadata associated with the
+// volume being published. Useful for modifying clients to make use of CSI
+// token requests.
+type ClientForMatadataFunc func(meta metadata.Metadata) (cmclient.Interface, error)
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -47,8 +47,13 @@ import (
 
 // Options used to construct a Manager
 type Options struct {
-	// Client used to interact with the cert-manager API
+	// Client is used to interact with the cert-manager API to list and delete
+	// requests.
 	Client cmclient.Interface
+
+	// CreateClient is used for returning a client that is used for creating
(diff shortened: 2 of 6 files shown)


JoshVanL:
Tested this and works how I want it to.

```yaml
spec:
  extra:
    authentication.kubernetes.io/pod-name:
    - my-csi-app-5d5bf8b8cd-nw2np
    authentication.kubernetes.io/pod-uid:
    - 38ae3ebf-9c7e-4e0e-bbdc-8fa3a7945c8f
  groups:
  - system:serviceaccounts
  - system:serviceaccounts:sandbox
  - system:authenticated
  uid: 743b8be6-b9cc-48f7-bb0c-1a4abfa86fb4
  username: system:serviceaccount:sandbox:my-csi-app
```

/unhold

storage/filesystem.go
```diff
@@ -165,6 +166,11 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 		return true, f.WriteMetadata(meta.VolumeID, meta)
 	}
 
+	if !apiequality.Semantic.DeepEqual(existingMeta.VolumeContext, meta.VolumeContext) {
+		f.log.WithValues("volume_id", meta.VolumeID).Info("volume context changed, updating file system metadata")
+		return true, f.WriteMetadata(meta.VolumeID, meta)
+	}
+
```
munnerz, line 173: I'm not sure I understand the purpose of this change - when/why could this metadata change? What case are we trying to handle here?

munnerz: Additionally: we should not overwrite the metadata file with `meta` (which likely won't contain any fields such as `nextIssuanceTime`) - instead, if we do want to support this, we should write `existingMetadata` back (after updating the context values). Otherwise we'll trigger an immediate re-issuance too.

JoshVanL: Thanks, I have added a commit which overwrites just the context to avoid re-issuance.

This change is needed since the Kubelet will [periodically call `NodePublishVolume` which renews any `TokenRequest` tokens](https://kubernetes-csi.github.io/docs/token-requests.html#usage). This is required so the driver always has a valid request token in the metadata when renewing a pods volume.

manager/manager.go
```diff
@@ -134,6 +144,7 @@ func NewManager(opts Options) (*Manager, error) {
 
 	m := &Manager{
 		client:         opts.Client,
```
munnerz, line 146: This `client` is not used anymore right? Because if `createClient` is not set in opts, it is being set to return `opts.Client`

JoshVanL: It's still used for deleting CRs: https://github.com/cert-manager/csi-lib/pull/9/files#diff-f679b61276b44dc3f4712950536707c0389ee4c536fc29837c0202c6e81c1bbfR358

JoshVanL:
```go
		if err := m.client.CertmanagerV1().CertificateRequests(toDelete.Namespace).Delete(ctx, toDelete.Name, metav1.DeleteOptions{}); err != nil {
```

manager/interfaces.go
```diff
@@ -82,3 +83,8 @@ type SignRequestFunc func(meta metadata.Metadata, key crypto.PrivateKey, request
 // The 'chain' and 'ca' arguments are PEM encoded and sourced directly from the
 // CertificateRequest, without any attempt to parse or decode the bytes.
 type WriteKeypairFunc func(meta metadata.Metadata, key crypto.PrivateKey, chain []byte, ca []byte) error
+
+// CreateClientFunc will return a cert-manager API client used for creating
+// objects. This is called with the metadata associated with the volume being
+// published. Useful for modifying clients to make use of CSI token requests.
+type CreateClientFunc func(meta metadata.Metadata) (cmclient.Interface, error)
```
munnerz, line 90: `ClientForMetadataFunc` instead perhaps? I think it is a small change, but sort of changes the implication from always _creating_ a client to also allow for 'returning' a client (i.e. if an implementation maintained a cache of clients).

Also - and this _isn't_ something to do here, but we should consider running codegen within this repository to avoid having this interface reference an externally generated clientset...

storage/filesystem.go
```diff
@@ -156,7 +157,7 @@ func (f *Filesystem) WriteMetadata(volumeID string, meta metadata.Metadata) erro
 }
 
 func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
```
munnerz, line 159: I think we also need to make a corresponding change to the 'in memory' storage implementation too (which is used during integration testing).

Adding to that, how can we write tests to verify this new functionality?

JoshVanL: Thanks, I have updated the inmemory store.

Any advise on how to test this functionality? I struggled to test this in `filesystem_test.go` for example.

PR #12, opened 2021-06-24
Allow storage implementations to optionally specifiy an fsGroup GID when writing files.

JoshVanL:
This PR changes the `DataWriter` interface to now accept metadata, rather than just the volume ID. The filesystem implementation now has the optional fields: 

```
	// FixedFSGroup is an optional field which will set the gid ownership of all
	// volume's data directories to this value.
	// If this value is set, FSGroupVolumeAttributeKey has no effect.
	FixedFSGroup *int64

	// FSGroupVolumeAttributeKey is an optional well-known key in the volume
	// attributes. If this attribute is present in the context when writing
	// files, gid ownership of the volume's data directory will be changed to
	// the value. Attribute value must be a valid int64 value.
	// If FixedFSGroup is defined, this field has no effect.
	FSGroupVolumeAttributeKey string
```

Filesystem will use these fields to optionally change the GID of the volume's data directory owner. 

This is a required addition since the Kubelet [will not](https://github.com/kubernetes/kubernetes/blob/6dd9deea3daa98e0003878544372b2c1b9e0e03d/pkg/volume/csi/csi_mounter.go#L382) `chown` CSI volumes that have the `readOnly` flag set, regardless of the CSI driver settings. The cert-manager CSI driver requires the `readOnly` flag is set to `true` for all mounted volumes.

We can also not discover the user and group of the container without making an API call to get the Pod spec. _We do not want to require CSI drivers to read Pods from the API._ There is an [accepted KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/2317-fsgroup-on-mount), which once implemented, will do away with the CSI driver needing this value passed and instead can rely on the volume context.

---

The intention for this value is that CSI driver implementations will either expose an option to users or hard code a GID value when calling `WriteFiles`. An istio CSR driver for example could hard code this value to `1337`.

The example implementation has been updated to expose a `csi.cert-manager.io/fs-group` volume attribute to set this value. Users will want to set this value to the same group context of the mounting container.

---

I'm not sure how we can reliability test this feature without e2e tests, as it requires `chown`ing.

### DIFF
--- a/storage/filesystem.go
+++ b/storage/filesystem.go
@@ -176,9 +176,20 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 	return false, nil
 }
 
-func (f *Filesystem) WriteFiles(volumeID string, files map[string][]byte) error {
-	if err := os.MkdirAll(f.dataPathForVolumeID(volumeID), 0644); err != nil {
+// WriteFiles writes the given data to filesystem files within the volume's
+// data directory. Filesystem supports setting custom fsUser UID to these
+// files.
+func (f *Filesystem) WriteFiles(volumeID string, files map[string][]byte, fsUser *int64) error {
+	// Data directory should be read, write and execute only to the fs user
+	if err := os.MkdirAll(f.dataPathForVolumeID(volumeID), 0700); err != nil {
 		return err
+	}
+
+	// If a fsUser is defined, Chown the directory to that user.
+	if fsUser != nil {
+		if err := os.Chown(f.dataPathForVolumeID(volumeID), int(*fsUser), -1); err != nil {
+			return fmt.Errorf("failed to chown data dir to uid %v: %w", *fsUser, err)
+		}
 	}
 
 	writer, err := util.NewAtomicWriter(f.dataPathForVolumeID(volumeID), fmt.Sprintf("volumeID %v", volumeID))
@@ -186,7 +197,7 @@ func (f *Filesystem) WriteFiles(volumeID string, files map[string][]byte) error 
 		return err
 	}
 
-	payload := makePayload(files)
+	payload := makePayload(files, fsUser)
 	return writer.Write(payload)
 }
 
@@ -223,13 +234,13 @@ func (f *Filesystem) tempfsPath() string {
 	return filepath.Join(f.baseDir, "inmemfs")
 }
 
-func makePayload(in map[string][]byte) map[string]util.FileProjection {
+func makePayload(in map[string][]byte, fsUser *int64) map[string]util.FileProjection {
 	out := make(map[string]util.FileProjection, len(in))
 	for name, data := range in {
 		out[name] = util.FileProjection{
-			Data: data,
-			// read-only for user + group (TODO: set fsUser/fsGroup)
-			Mode: readOnlyUserAndGroupFileMode,
+			Data:   data,
+			FsUser: fsUser,
+			Mode:   readOnlyUserAndGroupFileMode,
 		}
 	}
 	return out
--- a/storage/filesystem.go
+++ b/storage/filesystem.go
@@ -193,7 +193,7 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 // data directory. Filesystem supports changing ownership of the data directory
 // to a custom gid.
 func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte) error {
-	// Data directory should be read, write and execute only to the fs user; read and executable to group
+	// Data directory should be read and execute only to the fs user and group.
 	if err := os.MkdirAll(f.dataPathForVolumeID(meta.VolumeID), 0550); err != nil {
 		return err
 	}
@@ -220,17 +220,16 @@ func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte)
 		return err
 	}
 
-	// If a fsGroup is defined, Chown all files within the data directory.
+	// If a fsGroup is defined, Chown all files just written.
 	if fsGroup != nil {
-		dirName := f.dataPathForVolumeID(meta.VolumeID)
-		entries, err := os.ReadDir(dirName)
-		if err != nil {
-			return fmt.Errorf("failed to list files in data directory: %w", err)
-		}
+		// "..data" is the well-known location where the atomic writer links to the
+		// latest directory containing the files just written. Chown these real
+		// files.
+		dirName := filepath.Join(f.dataPathForVolumeID(meta.VolumeID), "..data")
 
-		for _, entry := range entries {
+		for filename := range files {
 			// Set the uid to -1 which means don't change ownership in Go.
-			if err := os.Chown(filepath.Join(dirName, entry.Name()), -1, int(*fsGroup)); err != nil {
+			if err := os.Chown(filepath.Join(dirName, filename), -1, int(*fsGroup)); err != nil {
 				return err
 			}
 		}
@@ -311,7 +310,7 @@ func (f *Filesystem) fsGroupForMetadata(meta metadata.Metadata) (*int64, error) 
 	// fsGroup has to be between 1 and 4294967295 inclusive. 4294967295 is the
 	// largest gid number on most modern operating systems. If the actual maximum
 	// is smaller on the running machine, then we will simply error later during
-	// the Chmod.
+	// the Chown.
 	if fsGroup <= 0 || fsGroup > 4294967295 {
 		return nil, fmt.Errorf("%q: gid value must be greater than 0 and less than 4294967295: %d", f.FSGroupVolumeAttributeKey, fsGroup)
 	}
--- a/storage/filesystem.go
+++ b/storage/filesystem.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/go-logr/logr"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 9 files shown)


storage/memory.go
```diff
@@ -126,7 +126,10 @@ func (m *MemoryFS) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 	return true, nil
 }
 
-func (m *MemoryFS) WriteFiles(volumeID string, files map[string][]byte) error {
+// WriteFiles writes the given data to virtual files in memory within the
+// volume's data directory. MemoryFS does not support setting user ownership of
+// files.
+func (m *MemoryFS) WriteFiles(volumeID string, files map[string][]byte, _ *int64) error {
```
JoshVanL, line 132: Should we return an error here if `fsUser != nil`?

munnerz:
The KEP seems to discuss `fsGroup` rather than `fsUser` - this PR seems to handle _both_. What's the reasoning here/should we be exposing controls around the owning _user_?

munnerz:
Upon closer inspection of the KEP, I'm inclined to say we should align (for now) with upstream and only exposing controlling the *group* (given that files are mounted in such a way that the group permissions will permit reading the files anyway)

JoshVanL:
> The KEP seems to discuss `fsGroup` rather than `fsUser` - this PR seems to handle _both_. What's the reasoning here/should we be exposing controls around the owning _user_?

This PR actually is only concerned with `fsUser`. The reason was that this is the only option for the `atomic_writer`. 

We could change this over to `fsGroup` instead, however the container will only be able to execute on the data directory, and not be able to read the data dir files. Will check this now.

JoshVanL:
As discussed, there is nothing stopping us adding fsGroup support to the atomic writer. This simplifies things a lot and keep us aligned.

third_party/util/atomic_writer.go
```diff
@@ -95,14 +96,14 @@ const (
 //      data to determine if an update is required.
 //  5.  A new timestamped dir is created
 //  6.  The payload is written to the new timestamped directory
-//  7.  Symlinks and directory for new user-visible files are created (if needed).
+//  7.  Symlinks and directory for new user-group-visible files are created (if needed).
```
munnerz, line 99: don't think this is an intentional change?

third_party/util/atomic_writer.go
```diff
@@ -394,19 +395,31 @@ func (w *AtomicWriter) writePayloadToDir(payload map[string]FileProjection, dir
...
 
-		if fileProjection.FsUser == nil {
+		// If neither FsUser or FsGroup defined, we don't need to Chown.
+		if fileProjection.FsUser == nil && fileProjection.FsGroup == nil {
 			continue
 		}
-		if err := os.Chown(fullPath, int(*fileProjection.FsUser), -1); err != nil {
-			klog.Errorf("%s: unable to change file %s with owner %v: %v", w.logContext, fullPath, int(*fileProjection.FsUser), err)
+
+		fsUser := -1
```
munnerz, line 403: what happens if we chown to -1? 🤔

storage/memory.go
JoshVanL: This is no longer relevant as we pass metadata  instead.

storage/filesystem.go
```diff
@@ -47,6 +48,18 @@ type Filesystem struct {
...
+	// volume's data directories to this value.
+	// If this value is set, FSGroupVolumeAttributeKey has no effect.
+	FixedFSGroup *int64
+
+	// FSGroupVolumeAttributeKey is an optional well-known key in the volume
+	// attributes. If this attribute is present in the context when writing
+	// files, gid ownership of the volume's data directory will be changed to
+	// the value. Attribute value must be a valid int64 value.
+	// If FixedFSGroup is defined, this field has no effect.
+	FSGroupVolumeAttributeKey *string
```
munnerz, line 62: I think we could make this a non-ptr and just check for it being empty? (as we don't need to disambiguate between the empty value and not set, unlike `FixedFSGroup`)

munnerz: (I only raise it because if this is a pointer, there is the possibility that someone could change it at runtime/midway through execution.. which is.. weird 😅)

storage/filesystem.go
```diff
@@ -176,17 +189,33 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
...
 }
 
-func (f *Filesystem) WriteFiles(volumeID string, files map[string][]byte) error {
-	if err := os.MkdirAll(f.dataPathForVolumeID(volumeID), 0644); err != nil {
+// WriteFiles writes the given data to filesystem files within the volume's
+// data directory. Filesystem supports changing ownership of the data directory
+// to a custom gid.
+func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte) error {
+	// Data directory should be read, write and execute only to the fs user; read and executable to group
+	if err := os.MkdirAll(f.dataPathForVolumeID(meta.VolumeID), 0750); err != nil {
```
munnerz, line 197: Hm, _do_ we want to permit writing by the owning user? Does this trip things up for the CSI driver if not? How come `0644` worked before, and now it needs to be `0750?`

munnerz: Presumably within the container this wouldn't be writable by anyone anyway due to it being a read-only FS. But do we need to keep it this way to allow the driver to write the data?

storage/filesystem.go
```diff
@@ -223,14 +252,42 @@ func (f *Filesystem) tempfsPath() string {
...
+		return nil, nil
+	}
+
+	fsGroupStr, ok := meta.VolumeContext[*f.FSGroupVolumeAttributeKey]
+	if !ok {
+		// If the attribute has not been set, return no ownership change.
+		return nil, nil
+	}
+
+	fsGroup, err := strconv.ParseInt(fsGroupStr, 10, 64)
```
munnerz, line 287: Should we also validate that the given gid is also a valid group ID? i.e. >0, perhaps less than some value too?

third_party/util/atomic_writer.go
JoshVanL: It was, but can change it back. The intention was that we _used_ to only have user-visible files in the sense that it only supported Chowning the user. We now support _user and group_.

third_party/util/atomic_writer.go
JoshVanL: This means the [ownership doesn't change](https://golang.org/src/os/file_posix.go?s=2985:3028#L97). Will add a comment.

storage/filesystem.go
JoshVanL: Fine by me

storage/filesystem.go
JoshVanL: Yep, I agree. Will add these check.

storage/filesystem.go
JoshVanL: Changed this to `0550` which does the job.

We need it to be readable and executable by both the user and group. 

The reason it worked before is that it wasn't run against a container with a different user/group.

storage/filesystem.go
```diff
@@ -176,18 +189,54 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 	return false, nil
 }
 
-func (f *Filesystem) WriteFiles(volumeID string, files map[string][]byte) error {
-	if err := os.MkdirAll(f.dataPathForVolumeID(volumeID), 0644); err != nil {
+// WriteFiles writes the given data to filesystem files within the volume's
+// data directory. Filesystem supports changing ownership of the data directory
+// to a custom gid.
+func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte) error {
+	// Data directory should be read, write and execute only to the fs user; read and executable to group
```
munnerz, line 196: comment needs updating :)

storage/filesystem.go
```diff
@@ -176,18 +189,54 @@ func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
...
 	payload := makePayload(files)
-	return writer.Write(payload)
+	if err := writer.Write(payload); err != nil {
+		return err
+	}
+
+	// If a fsGroup is defined, Chown all files within the data directory.
+	if fsGroup != nil {
+		dirName := f.dataPathForVolumeID(meta.VolumeID)
+		entries, err := os.ReadDir(dirName)
```
munnerz, line 226: Given the atomic writer creates subdirectories containing the actual files, does this work..? It seems like this would chown the directories containing data files, rather than the data itself?

JoshVanL: These are the files that get `Chowned` here:

```
csi-data-dir/inmemfs/csi-c3d3286f77ef69343fa88e568da8db465d919a403ba1daa27f628597b80533a9/data/..2021_06_25_15_49_43.932262773
csi-data-dir/inmemfs/csi-c3d3286f77ef69343fa88e568da8db465d919a403ba1daa27f628597b80533a9/data/..data
csi-data-dir/inmemfs/csi-c3d3286f77ef69343fa88e568da8db465d919a403ba1daa27f628597b80533a9/data/ca.crt
csi-data-dir/inmemfs/csi-c3d3286f77ef69343fa88e568da8db465d919a403ba1daa27f628597b80533a9/data/tls.crt
csi-data-dir/inmemfs/csi-c3d3286f77ef69343fa88e568da8db465d919a403ba1daa27f628597b80533a9/data/tls.key
```

Is this not correct/what we want?

JoshVanL: I have changed this to now Chmoding the real underlying files. We construct the path by using the atomic_writer's "..data" dir.

```go
	// If a fsGroup is defined, Chown all files just written.
	if fsGroup != nil {
		// "..data" is the well-known location where the atomic writer links to the
		// latest directory containing the files just written. Chown these real
		// files.
		dirName := filepath.Join(f.dataPathForVolumeID(meta.VolumeID), "..data")

		for filename := range files {
			// Set the uid to -1 which means don't change ownership in Go.
			if err := os.Chown(filepath.Join(dirName, filename), -1, int(*fsGroup)); err != nil {
				return err
			}
		}
	}
```

PR #17, opened 2021-08-11
Adds ClientForMetadataTokenRequestEmptyAud and EmptyAudienceTokenFromMetadata

JoshVanL:
This PR adds the `manager/util` package which is intended to include helper functions for csi-lib consumers.

`EmptyAudienceTokenFromMetadata` will return the empty audience request token from the volume context. This is useful for drivers who want to extract the Pod's ServiceAccount information without needed to GET Pods.

`ClientForMetadataTokenRequestEmptyAud` is intended to be plugged straight into `manager.ClientForMetadata` and will be used to generate cert-manager clients who authenticate to the API server using the Pod's ServiceAccount. 

I have ended up copying these funcs multiple times, so seems sensible to add these util helpers to csi-lib so other consumers don't have to write them themselves.

```release-note
NONE
```

/assign @munnerz

### DIFF
--- a/manager/util/tokenrequest.go
+++ b/manager/util/tokenrequest.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2021 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	cmclient "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
+	"k8s.io/client-go/rest"
+
+	"github.com/cert-manager/csi-lib/manager"
+	"github.com/cert-manager/csi-lib/metadata"
+)
+
+// ClientForMetadataTokenRequestEmptyAud returns a
+// manager.ClientForMetadataFunc that returns a cert-manager rest client whose
+// authentication is built using the passed empty audience ("") token request
+// in the metadata VolumeContext. The resulting cert-manager client is
+// authenticated against the Kubernetes API server using the mounting Pod's
+// ServiceAccount.
+//
+// Intended to be used as a manager ClientForMetadata so that created
+// CertificateRequests will have UserInfo fields of the mounting Pods
+// ServiceAccount.
+//
+// Drivers using this function _must_ have the empty audience tokenRequest
+// defined on the CSIDriver manifest definition, along with setting
+// requiresRepublish to true:
+//
+// tokenRequests:
+//   - audience: ""
+//     expirationSeconds: 3600
+// requiresRepublish: true
+//
+// restConfig must contain the Kubernetes API server Host, and a valid
+// TLSClientConfig.
+func ClientForMetadataTokenRequestEmptyAud(restConfig *rest.Config) manager.ClientForMetadataFunc {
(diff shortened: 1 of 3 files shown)


manager/util/tokenrequest.go
```diff
@@ -0,0 +1,123 @@
...
+	cmclient "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
+	"k8s.io/client-go/rest"
+
+	"github.com/cert-manager/csi-lib/manager"
+	"github.com/cert-manager/csi-lib/metadata"
+)
+
+// ClientForMetadataTokenRequestEmptyAud returns a
+// manager.ClientForMetadataFunc that returns a cert-manager rest client whose
+// authentication is built using the passed empty audience ("") token request
```
irbekrm, line 33: Suggestion: I understand that the empty audience is so that the audience defaults to kube API server. Would it make sense to add that to the comment or is it unlikely that a user of the function would not already know that?

manager/util/tokenrequest.go
```diff
@@ -0,0 +1,123 @@
...
+// Drivers using this function _must_ have the empty audience tokenRequest
+// defined on the CSIDriver manifest definition, along with setting
+// requiresRepublish to true:
+//
+// tokenRequests:
+//   - audience: ""
+//     expirationSeconds: 3600
+// requiresRepublish: true
+//
+// restConfig must contain the Kubernetes API server Host, and a valid
```
irbekrm, line 51: And I guess we validate those in `cmclient.NewForConfig` 👍🏼

manager/util/tokenrequest.go
```diff
@@ -0,0 +1,123 @@
...
+// TLSClientConfig.
+func ClientForMetadataTokenRequestEmptyAud(restConfig *rest.Config) manager.ClientForMetadataFunc {
```
irbekrm, line 53: Suggestion: indicate in the function name that it's cert-manager client.

irbekrm (approved):
Looks good to me from what I can tell - I have read through the code and tried to understand the context.

I guess once this gets merged, the functions would be used in `cert-manager-csi`? Then it might make sense to come back and add a link to where they are used.

I've added two optional suggestions, but feel free to unhold.
/hold
/lgtm

manager/util/tokenrequest.go
JoshVanL: I've tried to add this context in the following sentence:

> The resulting cert-manager client is
// authenticated against the Kubernetes API server using the mounting Pod's
// ServiceAccount.

It is not that it _defaults_ to the kube API server, it _is_ the kube API server audience, [see](https://kubernetes-csi.github.io/docs/token-requests.html#usage).

manager/util/tokenrequest.go
JoshVanL: We don't actually, but we can't really.. As a consumer, you get what you have given.

manager/util/tokenrequest.go
JoshVanL: I'm inclined to say no. All consumer facing clients are cert-manager clients, [see](https://github.com/cert-manager/csi-lib/blob/dd8ae5d66f5301940c865704f3e91992ab3bc4d6/manager/manager.go#L58).

I think extra context would be needed if it _wasn't_ a cert-manager client.

JoshVanL:
Thanks for the review @irbekrm :pray:

> I guess once this gets merged, the functions would be used in cert-manager-csi? Then it might make sense to come back and add a link to where they are used.

Not sure if that's needed too much. Ideally this will be used in multiple projects, I have a couple I need it for :smile:

manager/util/tokenrequest.go
irbekrm: Ah ok, that makes sense. I was thinking about it like about a controller where the default client think is a core kube client 👍🏼

manager/util/tokenrequest.go
irbekrm: Yeah makes sense especially since it's in the CSI dev docs too which I didn't realize 👍🏼

irbekrm:
Thanks for replying to all the comments!

/hold cancel

PR #19, opened 2022-01-25
Update to cert-manager 1.7

wallrj:
* Updated to cert-manager 1.7.1
* Update to Go 1.17 to work around some go mod problems with transitive dependencies
* Adjusted code that uses logr which now uses structs instead of interfaces.


example/main.go
```diff
@@ -112,7 +112,7 @@ func main() {
 			Client:             cmclient.NewForConfigOrDie(restConfig),
 			MetadataReader:     store,
 			Clock:              clock.RealClock{},
-			Log:                log,
+			Log:                &log,
```
wallrj, line 115: logr.Logger was an interface but now is a struct, so I've made it a pointer in the Options, so that we can tell when it has been set.

go.mod
```diff
@@ -1,19 +1,53 @@
...
+	google.golang.org/protobuf v1.27.1 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
+	k8s.io/api v0.23.1 // indirect
+	k8s.io/apiextensions-apiserver v0.23.1 // indirect
+	k8s.io/kube-aggregator v0.23.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
+	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
```
wallrj, line 50: This package requires Go 1.17
 * https://github.com/kubernetes-sigs/json/issues/8

test/log/test_logr.go
```diff
@@ -1,66 +0,0 @@
-/*
```
wallrj, line 1: There's a TestLogger in the go-logr package so I used that and  removed this.

wallrj (commented):
I'll need to update the testing repo to run these tests with Go 1.17

go.mod
```diff
@@ -1,19 +1,53 @@
 module github.com/cert-manager/csi-lib
 
-go 1.16
+go 1.17
```
munnerz, line 3: Is the minimum version of go required to use this library actually changing?

go.mod
```diff
@@ -1,19 +1,53 @@
...
 
 require (
 	github.com/container-storage-interface/spec v1.4.0
-	github.com/go-logr/logr v0.4.0
-	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/jetstack/cert-manager v1.3.1
+	github.com/go-logr/logr v1.2.0
+	github.com/jetstack/cert-manager v1.7.0-beta.0
```
munnerz, line 8: /hold until we have the stable version out 😊

test/util/testutil.go
```diff
@@ -118,7 +119,7 @@ func RunTestDriver(t *testing.T, opts DriverOptions) (DriverOptions, csi.NodeCli
 		WriteKeypair:         opts.WriteKeypair,
 	})
 
-	d := driver.NewWithListener(lis, opts.Log, driver.Options{
+	d := driver.NewWithListener(lis, *opts.Log, driver.Options{
```
munnerz, line 122: Hm, previously it would be possible for a user to pass a nil pointer here. Now it isn't (and a user would need to create a Discard logger). Would it make sense to take a pointer here instead, and if it's nil, instantiate a discard logger automatically?

go.mod
munnerz: Ah, saw your other comment - it seems like it is then 😊 I'm guessing 'go build ./...' won't succeed without this patch + without 1.17?

wallrj: Yeah, I've created https://github.com/jetstack/testing/pull/621 to update the go version used for testing.

manager/manager.go
```diff
@@ -150,7 +150,7 @@ func NewManager(opts Options) (*Manager, error) {
 		lister:            lister,
 		metadataReader:    opts.MetadataReader,
 		clock:             opts.Clock,
-		log:               opts.Log,
+		log:               *opts.Log,
```
munnerz, line 153: This will cause a panic if a user doesn't provide a logger (whereas the previous code would not). Can we initialise a `logr.Discard()` here to pass to the Manager if it is nil?

go.mod
wallrj: Done.

manager/manager.go
wallrj: There's already a check for nil `opts.Log` above on line 102 https://github.com/cert-manager/csi-lib/pull/19/files#diff-f679b61276b44dc3f4712950536707c0389ee4c536fc29837c0202c6e81c1bbfR102

test/util/testutil.go
wallrj: That's already handled above, right? We create a TestLogger if `opts.Log` is nil.
I will leave it like this for now and you can improve it in a followup PR if necessary.

PR #21, opened 2022-03-31
storage: create data directory when RegisterMetadata is called

munnerz:
This fixes a bug in the recently merged continue on not ready functionality (#20).

Previously we'd only create the `data` directory within each volume's directory in the tmpfs when `WriteFiles` is called. Because `WriteFiles` is not called if the volume is not ready to be issued for, this directory never existed and so the `data` directory couldn't be mounted into the pod when `Mount` was called.

By moving the creation of the full directory structure to the point where RegisterMetadata is called, we then allow the volume's data directory to be mounted even if no data has been written to it yet.

/cc @JoshVanL @7ing


JoshVanL:
Nasty!

This change makes sense to me. Thanks @munnerz 

Let's publish a patch release when this is merged

/lgtm

Issue #24, opened 2022-05-17
[bug] labelSelectorForVolume fail to add labels


7ing:
We notice there is only one CertificateRequest object exists for a pod with two CSI volumes.
From the watch log, it seems we had two CSRs, but one got purged somehow.

The root cause is:
https://github.com/cert-manager/csi-lib/blob/2f2e205d5f03a4dbc089c766d49888e42e614fc6/manager/manager.go#L391
```
sel := labels.NewSelector()
req, err := labels.NewRequirement(...)
sel.Add(*req) // this won't add req to the selectors... 
```

simple fix is like:

```
sel = sel.Add(*req)
```

PR #25, opened 2022-05-17
bug fix: labelSelectorForVolume

7ing:
labelSelector should add req and assgin back.

fixes: #24


go.mod
```diff
@@ -18,19 +18,56 @@ require (
 )
 
 require (
+	github.com/beorn7/perks v1.0.1 // indirect
```
7ing, line 21: auto generated by `go mod tidy -compat=1.17`

munnerz: I think these changes can be excluded from this PR, especially as our presubmits run with Go 1.18 now: https://github.com/jetstack/testing/blob/3bb9af8d51d14f32f6d3961a66e47e370d4d537e/config/jobs/cert-manager/csi-lib/cert-manager-csi-lib-presubmits.yaml#L16

(plus also exclude changes to go.sum as well)

Unless these came about due to the new tests you've added in this PR too? 👀

7ing: Yes, these changes are due to new tests...required during `go test ./...`

PR #28, opened 2022-07-13
Include Pod UID on CertificateRequest resources

davidsbond:
This commit adds an additional label to the generated CertificateRequest
resources that includes the UID of the Pod that initiated the request via
the csi-driver implementation. The Pod UID is taken from the volume context
in the same way that is used to created the owner references.

Adding this label adds increased queryablility to see which pod generated
which certificate request. Internally at Jetstack, we want a way to monitor
unused certificates, so this label will help us in checking the specific pod.

It could be that some third party tool adds additional owner references to the
CertificateRequest resource, so this label (providing it isn't changed by another
third party) will give insight into the specific pod making the request.

Related to https://github.com/cert-manager/csi-driver/pull/102

Signed-off-by: David Bond <davidsbond93@gmail.com>

### DIFF
--- a/internal/api/consts.go
+++ b/internal/api/consts.go
@@ -19,4 +19,5 @@ package api
 const (
 	NodeIDHashLabelKey   = "csi.cert-manager.io/node-id-hash"
 	VolumeIDHashLabelKey = "csi.cert-manager.io/volume-id-hash"
+	PodUIDLabelKey       = "csi.cert-manager.io/pod-uid"
 )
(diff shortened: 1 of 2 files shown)


JoshVanL:
Thanks @davidsbond 

/lgtm
/approve
/ok-to-test

/hold
In case you have any objections @munnerz

JoshVanL:
/lgtm cancel

@davidsbond, on second thoughts, could you please add an integration test for this?

munnerz:
@davidsbond are you still interesting in getting this merged? Can you add a little test as Josh mentioned?

PR #32, opened 2022-09-06
Use a passed through context to manager issue function

JoshVanL:
It is not clear to me why a TODO context is being used for the issuing function. It seems clear to me that we would want to use the context which is cancelled when the stop channel is closed.

PR passes that context through to `issue`.

/assign @munnerz

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -243,8 +243,7 @@ type Manager struct {
 }
 
 // issue will step through the entire issuance flow for a volume.
-func (m *Manager) issue(volumeID string) error {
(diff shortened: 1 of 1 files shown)


7ing (approved):
lgtm

munnerz:
👍 I think this may be an artefact from when the manager was first written, and how `issue` was invoked wasn't clear (i.e. would we want to have some per-issuance-attempt context deadline set as well as the overall stopCh one).

That doesn't need to block progress though, and this is progress 😄 

/lgtm

PR #34, opened 2022-09-07
E2E testing boilerplate

JoshVanL:
This PR adds the initial e2e test boilerplate code.

The script `hack/run-e2e.sh` spins up a kind cluster, loads the example cert-manager-csi conainer, installs cert-manager+csi-driver then runs the e2e test suite.

There is currently only a dummy TODO test written.

The PR experiments with using nix flakes for dependency management and building the project. You should invoke the test script using `nix develop -c ./hack/run-e2e.sh` (once you have installed nix first!).

We need to build kind ourselves since that is not available in the upstream nixpkgs.

/cc @munnerz 
/hold

Waiting to get a nix CI image released so we can see how this runs in prow.


Dockerfile
```diff
@@ -1,21 +0,0 @@
-# Copyright 2021 The cert-manager Authors.
```
JoshVanL, line 1: File is no longer needed as it has been moved to `flake.nix` as an output.

flake.nix
```diff
@@ -0,0 +1,131 @@
...
+            homepage = "https://github.com/kubernetes-sigs/kind";
+            description = "Kubernetes IN Docker - local clusters for testing Kubernetes";
+            longDescription = ''
+              kind is a tool for running local Kubernetes clusters using Docker
+              container "nodes". kind was primarily designed for testing Kubernetes
+              itself, but may be used for local development or CI.
+            '';
+            license = licenses.asl20;
+          };
+        };
```
JoshVanL, line 95: kind is not available in nixpkgs.

flake.nix
```diff
@@ -0,0 +1,131 @@
...
+          inherit src vendorSha256;
+          buildPhase = ''
+            go test -v --race -o csi-lib-e2e -c ./test/e2e/.
+          '';
+          postInstall = ''
+            mkdir -p $out/bin
+            mv csi-lib-e2e $out/bin/.
+          '';
+          doCheck = false;
+        };
```
JoshVanL, line 64: We build the csi-lib e2e binary.

hack/run-e2e.sh
```diff
@@ -0,0 +1,71 @@
...
+CERT_MANAGER_VERSION="${CERT_MANAGER_VERSION:-v1.9.1}"
+
+ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
+cd "$ROOT_DIR"
+
+check_command() {
+  if ! [ -x "$(command -v $1)" ]; then
+    echo "Error: $1 is not installed." >&2
+    echo "Invoke script using 'nix develop -c ./hack/run.sh'" >&2
+    echo "https://nixos.org/download.html" >&2
```
JoshVanL, line 31: !

hack/run-e2e.sh
```diff
@@ -66,6 +66,7 @@ kind load image-archive --name cert-manager-csi-e2e $(nix build --print-out-path
 
 echo "> Loading csi-lib docker image..."
 kind load image-archive --name cert-manager-csi-e2e <(gzip --decompress --stdout $(nix build --print-out-paths .#container)) &
+kind load image-archive --name cert-manager-csi-e2e $(nix build --print-out-paths .#csi-node-driver-registrar-image) &
```
munnerz, line 69: This is neat (although looks weird in diff output due to being interpreted as a comment.. does adding quote marks around the argument make it render better?

.gitignore
```diff
@@ -0,0 +1 @@
+/result
```
munnerz, line 1: What is `/result`? Can you add a comment?

hack/run-e2e.sh
```diff
@@ -0,0 +1,86 @@
...
+ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
+cd "$ROOT_DIR"
+
+TMP_DIR=$(mktemp -d)
+trap "rm -rf $TMP_DIR" EXIT
+
+check_command() {
+  if ! [ -x "$(command -v $1)" ]; then
+    echo "Error: $1 is not installed." >&2
+    echo "Invoke script using 'nix develop -c ./hack/run.sh'" >&2
```
munnerz, line 31: Can we automatically invoke `nix develop -c ./hack/run-e2e.sh` if a user calls this script directly without `nix` (assuming that it is always required). We can check `$0` to determine how the script was invoked :)

munnerz: Also `s/run.sh/run-e2e.sh/`

go.mod
```diff
@@ -7,8 +7,10 @@ require (
 	github.com/container-storage-interface/spec v1.6.0
 	github.com/go-logr/logr v1.2.3
 	github.com/kubernetes-csi/csi-lib-utils v0.11.0
+	github.com/onsi/ginkgo/v2 v2.1.6
+	github.com/onsi/gomega v1.20.1
```
munnerz, line 11: It's a shame that we have to pull in these test dependencies in our main go.mod file... Not sure what the best strategies are for avoiding this, but it could well make things harder in future with consumers of this library that e.g. depend on `controller-runtime` to provide a Client (or even depend on cert-manager which itself depends on these libraries)

hack/nix/overlay-kind.nix
```diff
@@ -0,0 +1,18 @@
+final: prev: {
+  # Overlay kind to version v0.15.0
```
munnerz, line 2: Hmm.. this looks like an interesting file, but I'm not totally sure what it does or how it works 👀 can you add some comments, and possibly links to any relevant articles to help understand some of this?

flake.lock
```diff
@@ -0,0 +1,42 @@
+{
```
munnerz, line 1: How do I update this file?

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; overlays = [ (import ./hack/nix/overlay-kind.nix) ]; };
+
+        # We only source go files to have better cache hits when actively
+        # working on non-go files.
+        src = pkgs.lib.sourceFilesBySuffices ./. [ ".go" "go.mod" "go.sum" ];
```
munnerz, line 35: What does this do? If we depend on say, a yaml file or some other testdata in our tests, does this not match them/what does it do? 👀

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+        vendorSha256 = "sha256-3pNKmR+yKIf/15eftJyHD7m7LerFbZ2m+N6zxVXz2sU=";
```
munnerz, line 36: How do we update this/what is it? 👀

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+          name = "csi-lib-e2e";
+          inherit src vendorSha256;
+          buildPhase = ''
+            go test -v --race -o csi-lib-e2e -c ./test/e2e/.
+          '';
+          postInstall = ''
+            mkdir -p $out/bin
+            mv csi-lib-e2e $out/bin/.
+          '';
+          doCheck = false;
```
munnerz, line 78: What is `doCheck` and why is it disabled here but not mentioned in the `cert-manager-csi` definition?

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+        };
+
+        containerImage = pkgs.dockerTools.buildImage {
+          name = "cert-manager-csi";
+          tag = "example";
+          # mount and umount are required for csi-driver functions.
```
munnerz, line 84: Where is this being copied from? We didn't do this previously in our Dockerfile right? 👀

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+        csi-node-driver-registrar-image = pkgs.dockerTools.pullImage{
+          imageName = "quay.io/k8scsi/csi-node-driver-registrar";
+          imageDigest = csi-node-driver-registrar-digest;
+          sha256 = csi-node-driver-registrar-sha256;
+          finalImageTag = "v${csi-node-driver-registrar-version}";
+          finalImageName = "quay.io/k8scsi/csi-node-driver-registrar";
+        };
+
+        busybox-image = pkgs.dockerTools.pullImage{
+          imageName = "busybox";
```
munnerz, line 150: What's the reasoning to switching to busybox and manually copying in binaries from the host/build environment rather than using alpine and `apk` to install the required tools? Does nix allow running arbitrary commands whilst building docker images or does it prevent it to ensure a hermetic build?

flake.nix
```diff
@@ -0,0 +1,184 @@
...
+          inherit cert-manager-controller-image;
+          inherit cert-manager-webhook-image;
+          inherit cert-manager-cainjector-image;
+          inherit cert-manager-ctl-image;
+          inherit cert-manager-csi;
+          inherit busybox-image;
+          inherit csi-node-driver-registrar-image;
+          inherit csi-lib-e2e;
+        };
+        devShells.default = pkgs.mkShell {
```
munnerz, line 172: What's a devShell? 😄 looks like it is provisioning a local development environment? Can I run this on my local machine? How does it work across macOS vs Linux etc?

munnerz (commented):
Thanks for putting this together - I've asked quite a few questions (a lot of which are due to my lack of understanding of nix build rules etc).

I think it'd be great to see more of this documented/explained, as you'll otherwise find yourself as the sole-maintainer of these files 😅.

I think I can see the benefits of a system like this, especially if we consider we have to maintain these sorts of build definitions that are utilised across multiple repositories, but with the way it's currently put together here it seems like we'll be repeating these definitions between different repos again anyway. Is there any way we can create some form of common `cert-manager/nix-pkgs` repository, and then have a slimmer/less complex development/test environment setup contributed here? We could also start out by rolling this out as a _development environment_ too, which will help others to become more familiar with nix for package, build and dependency management.

.gitignore
JoshVanL: `result` is the default location of where `nix build` writes to.

hack/run-e2e.sh
JoshVanL: There is not a great way to determine whether the script has been invoked by `nix develop -c hack/run-e2e.sh` when _inside_ the script.

Best we could do is test an env var which is used when a shell is setup from nix (`IN_NIX_SHELL`). If this variable isn't present, then we should be okay to include `exec nix develop -c "${ROOT_DIR}/hack/run-e2e.sh"` in the script.

go.mod
JoshVanL: I've gone ahead and created a `go.mod` under `/test/e2e`. This seems to be working quite well.

hack/nix/overlay-kind.nix
JoshVanL: Sure thing!

To share here also: this file contains a nix ["overlay"](https://nixos.wiki/wiki/Overlays). Overlays are the canonical method in nix to overwrite definitions. In this case we are overriding kind from version `v0.14.0` to `v1.15.0`. This is a good way for using newer versions than that in nixpkgs or adding patches etc.

As I wrote this message I see that kind has now been updated to v0.15.0 upstream so this overlay can be removed.

flake.nix
```diff
@@ -0,0 +1,183 @@
...
+          buildInputs = [
+            pkgs.kubectl
```
munnerz, line 173: How are these versioned/how do we select which version to use? It's currently pulling in a v1.23 binary but I don't see us specifying that anywhere

flake.nix
JoshVanL: This gets all the files under `./.` with the name prefix `.go`, `go.mod`, and `go.sum` (all go files).

That is correct- if a test depended on a non-file then we would need to use a different src strategy or not bother with the filtering at all if we didn't think it was worth it.

flake.lock
JoshVanL: `$ nix flake update`

flake.nix
JoshVanL: There do exist nix [`prefetch`](https://nixos.org/manual/nix/stable/command-ref/nix-prefetch-url.html) utilities for getting these shas, however I have always found it easier to just run `nix develop` etc. and copy+paste when I get an errors :smile:

flake.nix
JoshVanL: Not sure why `doCheck` was included. Think that was a copy+paste error.

flake.nix
JoshVanL: Correct, in the existing Dockerfile we used alpine as the base image which had `mount` and `umount` installed. Instead, we now use the mount packages from nixpkgs.

flake.nix
JoshVanL: We are actually pulling busybox here as that is the example application we use in `/deploy/example/example-app.yaml`. This basically means we can load the image in, rather than it being downloaded every time inside kind. This speeds up tests with a warm cache.

AFAIK nix also does not allow you to run arbitrary commands when building images.

flake.nix
JoshVanL: Correct, this is setting up your local environment in your shell. You can access that shell with `nix develop`. 

This should work on all systems, so long as they are included in ` flake-utils.lib.eachDefaultSystem `.

```
$ flake show
git+file:///keep/home/go/src/github.com/cert-manager/csi-lib?ref=refs%2fheads%2fe2e-nix-boilerplate&rev=17ae38eb12de438350b90fdc8e79e15f9362090f
├───devShells
│   ├───aarch64-darwin
│   │   └───default: development environment 'nix-shell'
│   ├───aarch64-linux
│   │   └───default: development environment 'nix-shell'
│   ├───i686-linux
│   │   └───default: development environment 'nix-shell'
│   ├───x86_64-darwin
│   │   └───default: development environment 'nix-shell'
│   └───x86_64-linux
│       └───default: development environment 'nix-shell'
```

inteon:
Superseded by https://github.com/cert-manager/csi-lib/pull/75

Issue #30, opened 2022-08-23
CertificateRequest should be in terminate state when it is denied by an approver


7ing:
**Related code block**: 
https://github.com/cert-manager/csi-lib/blob/main/manager/manager.go#L305-L309
```golang
// Handle cases where the request has been explicitly denied
if apiutil.CertificateRequestIsDenied(updatedReq) {
  cond := apiutil.GetCertificateRequestCondition(updatedReq, cmapi.CertificateRequestConditionDenied)
  return false, fmt.Errorf("request %q has been denied by the approval plugin: %s", updatedReq.Name, cond.Message)
}
```

**Observed behavior**:
When `CertificateRequest` is denied by an approver, the lib will delete it and recreate one. If the approver is auto approve/deny the CSR based on some policies, this would create an infinite loop: create -> denied -> delete -> create again -> ...

**Sample logs**:
```
....
I0823 16:00:25.368111       1 nodeserver.go:76] driver "msg"="Registered new volume with storage backend" "pod_name"="csi-64d84b77c-vfblf"
I0823 16:00:25.368202       1 nodeserver.go:86] driver "msg"="Volume registered for management" "pod_name"="csi-64d84b77c-vfblf"
I0823 16:00:25.368376       1 nodeserver.go:97] driver "msg"="Waiting for certificate to be issued..." "pod_name"="csi-64d84b77c-vfblf"
I0823 16:00:26.368756       1 manager.go:514] manager "msg"="Triggering new issuance" "volume_id"="csi-1d842b8307fb94812b128fc0dad4284e119e757ca39bbc164a6a693e8cc2cb80"
I0823 16:00:26.368811       1 manager.go:249] manager "msg"="Processing issuance" "volume_id"="csi-1d842b8307fb94812b128fc0dad4284e119e757ca39bbc164a6a693e8cc2cb80"
I0823 16:00:26.385125       1 manager.go:379] manager "msg"="Deleted CertificateRequest resource" "name"="c7722b4d-cab5-4709-99ed-1576fd4f705e" "namespace"="poc" "volume_id"="csi-1d842b8307fb94812b128fc0dad4284e119e757ca39bbc164a6a693e8cc2cb80"
I0823 16:00:26.503681       1 manager.go:287] manager "msg"="Created new CertificateRequest resource" "volume_id"="csi-1d842b8307fb94812b128fc0dad4284e119e757ca39bbc164a6a693e8cc2cb80"
[...]

PR #35, opened 2022-09-09
Simplify issuance backoff handling & improve error messages

munnerz:
This PR changes the internals of how the NodePublishVolume calls wait for an initial issuance of a Certificate.

Previously, a volume would be registered for management in the NodePublishVolume call, and then a `wait` function was used from the driver package to wait until the manager has successfully completed an issuance of the certificate.

This made it take longer for users to see errors if there are any (it'd always take a minimum of 30s to get errors) plus it has lead to poor handling of "backing off" in cases where requests continuously fail (see #30).

Instead, the driver will now attempt *one* issuance prior to beginning the long-running renewal loop in the manager, by using the new `ManageVolumeImmediate` method. This method:

1) registers the volume as under management (to prevent any other NodePublishVolume calls that run at the same time for the volume duplicating work/interfering) but does NOT start the renewal loop
2) makes a call to `issue` to attempt a single issuance
3) if this fails, it returns so the driver can call *unmanage* and clean up volume data on disk
4) if it succeeds, it begins the renewal goroutine like usual

This allows us to lean on the kubelet/CSI driver consumer  to define the retry behaviour if a pod isn't able to start.

In cases where a *renewal* is failing, the same exponential backoff behaviour we had before will be applied (and the exponent will not be thrown away every 30s, like described in #30).

I've also tuned the exponential backoff parameters so the initial 'base' backoff in these cases is 30s, with a factor of 2 and a max cap of 5 minutes. This leads to a pattern of waiting T seconds after each attempt:

```
T=0s
T=30s
T=60s
T=120s
T=240s
T=300s (5 minutes, the cap)
T=300s (5 minutes, the cap)
...
```

This backoff behaviour/configuration can now also be customised by library consumers to their liking.

As part of this, I've also improved the error messages returned to users in cases where their certificate requests are not being approved or issued (instead of just showing `timed out waiting for the condition`).

cc @7ing @JoshVanL

### DIFF
--- a/driver/nodeserver.go
+++ b/driver/nodeserver.go
@@ -26,7 +26,6 @@ import (
 	"github.com/go-logr/logr"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/mount-utils"
 
 	"github.com/cert-manager/csi-lib/manager"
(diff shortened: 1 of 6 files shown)


munnerz:
To note as well, the kubelet's standard backoff handling also appears to be exponential. We will rely on the kubelet to handle/time retries if a pod is in a ContainerCreating state - our own exponential backoff only comes into effect for *renewals*.

We may want to consider surfacing failed renewals via metrics in the CSI driver or something too, to make it easier for operators to identify 'stuck' renewals more easily.

driver/nodeserver.go
```diff
@@ -48,7 +47,7 @@ type nodeServer struct {
 func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
 	meta := metadata.FromNodePublishVolumeRequest(req)
 	log := loggerForMetadata(ns.log, meta)
-	ctx, _ = context.WithTimeout(ctx, time.Second*30)
+	ctx, _ = context.WithTimeout(ctx, time.Second*60)
```
munnerz, line 50: I've increased this as we now return errors from approvers/issuers as soon as they happen, which makes it easier for us to tolerate issuers that take longer (as it's no longer at the cost of user experience when a failure is happening and they have to wait until this time/cap before seeing an error)

driver/nodeserver.go
```diff
@@ -48,7 +47,7 @@ type nodeServer struct {
...
```
JoshVanL, line 51: We should probably be cancelling this context on return.

```suggestion
	ctx, cancel = context.WithTimeout(ctx, time.Second*60)
        defer cancel()
```

```

File: manager/manager.go [hunk 3]
```diff
@@ -82,6 +82,9 @@ type Options struct {
 	SignRequest        SignRequestFunc
 	WriteKeypair       WriteKeypairFunc
 	ReadyToRequest     ReadyToRequestFunc
+
+	// BackoffConfig configures the exponential backoff applied to certificate renewal failures.
+	BackoffConfig *wait.Backoff
```
JoshVanL, line 87: This is more explicit. I would always have to refer the field comment without `Renewal` prefix. 

```suggestion
	// RenewalBackoffConfig configures the exponential backoff applied to certificate renewal failures.
	RenewalBackoffConfig *wait.Backoff
```

manager/manager.go
```diff
@@ -304,11 +351,25 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
 		// Handle cases where the request has been explicitly denied
 		if apiutil.CertificateRequestIsDenied(updatedReq) {
 			cond := apiutil.GetCertificateRequestCondition(updatedReq, cmapi.CertificateRequestConditionDenied)
+			// if a CR has been explicitly denied, we DO stop execution.
+			// there may be a case to be made that we could continue anyway even if the issuer ignores the approval
+			// status, however these cases are likely few and far between and this makes denial more responsive.
```
JoshVanL, line 356: :+1:

manager/manager.go
```diff
@@ -454,22 +521,67 @@ func (m *Manager) submitRequest(ctx context.Context, meta metadata.Metadata, csr
 	return req, nil
 }
 
-// ManageVolume will initiate management of data for the given volumeID.
-func (m *Manager) ManageVolume(volumeID string) error {
+// ManageVolumeImmediate will register a volume for management and immediately attempt a single issuance.
+// This
```
JoshVanL, line 525: Missing rest of sentence :smile:

JoshVanL (changes requested):
Thanks @munnerz, couple comments. I have walked through the state machine and the changes make sense to me.

PR #36, opened 2022-09-13
Default renewal time to halfway through certificates lifetime

munnerz:
Since #35, we now rely on the next issuance time being set to properly 'resume' monitoring certificates after a driver restart.

Whilst we have always informed implementors to set the NextIssuanceTime field during their own `writeKeypair` implementation, we don't actually set this field in the core library.

This PR changes the issuance routine to explicitly default the NextIssuanceTime to be halfway through a certificate's lifetime. Implementors are free to override this default in their own `writeKeypair` function (and likely will/do). This serves as a default to avoid invalid/broken states from occurring (e.g. a certificate has been fully issued but the nextIssuanceTime is never set).

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -18,6 +18,8 @@ package manager
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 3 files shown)


manager/manager.go
```diff
@@ -336,6 +338,18 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
...
+	// Default the renewal time to be halfway through the certificate's duration.
+	// The implementation's writeKeypair function may override this value before
+	// writing to the storage layer.
+	block, _ := pem.Decode(req.Status.Certificate)
+	crt, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("parsing issued certificate: %w", err)
+	}
+	duration := crt.NotAfter.Sub(crt.NotBefore)
+	midpoint := crt.NotBefore.Add(duration / 2)
```
JoshVanL, line 350: This is different to other implementations across cert-manager, which default to `2/3`rds. Is the rational that csi-lib certs are expected to be shorter lived so there is less wriggle room during renewal?

munnerz: There wasn't particularly clear rationale for 1/2 vs 1/3rd nope. In pretty much all cases this _should_ be set by the `writeKeypair` function, so this should always be overridden. This is more of a safety net.. I'm pretty indifferent either way as to what we default this to, so can change it over to 2/3rds of the way through its life if preferred :)

JoshVanL: Yes please- just to have consistency and rule of least surprise.

munnerz: Done

PR #42, opened 2022-10-21
Make NodePublishVolume blocking if a volume is already ready to request

### LINKED ISSUE #41: Expect ManageVolumeImmediate() is a blocking call that waits for certificate to be issued
`ManageVolumeImmediate()` is introduced in #35. After this PR, we observed **inconsistent** behaviors in following settings:
- Driver's `ContinueOnNotReady` is set to `true`
- Manager's `ReadyToRequestFunc` is set as `AlwaysReadyToRequest`

### **Before**
**New pod won't come up until the certificate is issued / written to host path.** 
Relevant code in `v0.3.0`:
https://github.com/cert-manager/csi-lib/blob/8e344a2d37de27bacb17f8e1d5ddc1f03733351a/driver/nodeserver.go#L96-L102

### **Now**
**New pod can come up before the certificate is issued / written to host path.** 
Relevant code in `main`:
1. `meta.NextIssuanceTime` is set to epoch time.
https://github.com/cert-manager/csi-lib/blob/7dfb05b4437c66ee38004f725ec9b7be80d2280a/driver/nodeserver.go#L71-L77
2. `ManageVolumeImmediate` is called.
https://github.com/cert-manager/csi-lib/blob/7dfb05b4437c66ee38004f725ec9b7be80d2280a/driver/nodeserver.go#L95-L100
3. cert is issued asynchronously in `startRenewalRoutine` call.
https://github.com/cert-manager/csi-lib/blob/7dfb05b4437c66ee38004f725ec9b7be80d2280a/manager/manager.go#L559-L570

### Thoughts
In most cases, this won't be an issue at all. But this inconsistent behavior may break some deployments that don't have checks on cert files (e.g., using init-container). Sample manifest that would get impact on this:
```yaml
...
spec:
  containers:
  - name: sample-app
     ...
     volumeMounts:
     - mountPath: /path/to/file/tls.crt
        name: csi-cert
        subPath: tls.crt
  volumes:
  - csi:
        driver: sample.csi.driver
        readOnly: true
        name: csi-cert
...
```
Since kubelet cannot find the file, it will throw following error and prevent container from starting up:
```yaml
state:
      waiting:
        message: 'failed to start container "ca79b3d8da42ce7f2d92409d0e52260005b1b02a1231997ad3679db95358128f":
          Error response from daemon: failed to create shim task: OCI runtime create
          failed: runc create failed: unable to start container process: error during
          container init: error mounting "/var/lib/kubelet/pods/449f2b9e-e0c5-4781-8177-626984f324bd/volume-subpaths/csi-cert/sample-app/1"
          to rootfs at "/path/to/file/tls.crt": mount /var/lib/kubelet/pods/449f2b9e-e0c5-4781-8177-626984f324bd/volume-subpaths/csi-cert/sample-app/1:/path/to/file/tls.crt
          (via /proc/self/fd/6), flags: 0x5001: not a directory: unknown: Are you
          trying to mount a directory onto a file (or vice-versa)? Check if the specified
          host path exists and is the expected type'
        reason: RunContainerError
```

munnerz:
Previously if the NextIssuanceTime was set to the epoch, ManageVolumeImmediate wouldn't actually block.

Fixes #41 

cc @7ing @JoshVanL

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -557,7 +557,7 @@ func (m *Manager) ManageVolumeImmediate(ctx context.Context, volumeID string) (m
 	}
 
 	// Only attempt issuance immediately if there isn't already an issued certificate
-	if meta.NextIssuanceTime == nil {
+	if meta.NextIssuanceTime == nil || meta.NextIssuanceTime.IsZero() {
 		// If issuance fails, immediately return without retrying so the caller can decide
 		// how to proceed depending on the context this method was called within.
 		if err := m.issue(ctx, volumeID); err != nil {
--- a/test/integration/ready_to_request_test.go
+++ b/test/integration/ready_to_request_test.go
@@ -40,6 +40,81 @@ import (
 	testutil "github.com/cert-manager/csi-lib/test/util"
 )
 
+// Tests to ensure that if a certificate is immediately ready to be requested when NodePublishVolume
+// is called, the call will be blocking until the volume is actually ready.
+func Test_PublishCallBlocksIfReadyToRequestDespiteContinueOnNotReadyTrue(t *testing.T) {
+	store := storage.NewMemoryFS()
+	clock := fakeclock.NewFakeClock(time.Now())
+
+	opts, cl, stop := testdriver.Run(t, testdriver.Options{
+		Store:              store,
+		Clock:              clock,
+		ContinueOnNotReady: true,
+		ReadyToRequest:     manager.AlwaysReadyToRequest,
+		GeneratePrivateKey: func(meta metadata.Metadata) (crypto.PrivateKey, error) {
+			return nil, nil
+		},
+		GenerateRequest: func(meta metadata.Metadata) (*manager.CertificateRequestBundle, error) {
+			return &manager.CertificateRequestBundle{
+				Namespace: "certificaterequest-namespace",
+			}, nil
+		},
+		SignRequest: func(meta metadata.Metadata, key crypto.PrivateKey, request *x509.CertificateRequest) (csr []byte, err error) {
+			return []byte{}, nil
+		},
+		WriteKeypair: func(meta metadata.Metadata, key crypto.PrivateKey, chain []byte, ca []byte) error {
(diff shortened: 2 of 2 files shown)


test/integration/ready_to_request_test.go
```diff
@@ -40,6 +40,78 @@ import (
...
+	defer close(stopCh)
+
+	tmpDir, err := os.MkdirTemp("", "*")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+	_, err = cl.NodePublishVolume(ctx, &csi.NodePublishVolumeRequest{
```
JoshVanL, line 89: Could you add a comment here saying that this is the function that is going to block, then another above the read file below saying that we can immediately check because it blocks.

The test wasn't super clear the first time I read through it.

manager/manager.go
```diff
@@ -556,8 +556,9 @@ func (m *Manager) ManageVolumeImmediate(ctx context.Context, volumeID string) (m
 		return true, fmt.Errorf("reading metadata: %w", err)
 	}
 
+	epoch := time.Time{}
 	// Only attempt issuance immediately if there isn't already an issued certificate
-	if meta.NextIssuanceTime == nil {
+	if meta.NextIssuanceTime == nil || *meta.NextIssuanceTime == epoch {
```
JoshVanL, line 561: (*meta.NextIssuanceTime).IsZero()

JoshVanL (changes requested):
Makes sense 👍

7ing (approved):
lgtm, thanks @munnerz

test/integration/ready_to_request_test.go
munnerz: Done

manager/manager.go
munnerz: Done

Issue #45, opened 2023-02-24
Exponential backoff handling does not apply to certificate renewal in pending phase


7ing:
# Problem
When a certificate get renewal, the driver will generate a new `CertificateRequest` and wait for it being ready. If there is an error, the driver will do a retry in an exponential backoff pattern. This feature is brought by https://github.com/cert-manager/csi-lib/pull/35.
However, if the `CertificateRequest` is in a pending state (e.g., not being handled by CSR controller, not approved, or approved but not issued yet, etc.), **the driver will wait infinitely instead of exponential backoff retry**. This behavior is not inline with the above failure situation. 

# Reproduce
We could reproduce it by faking a long waiting work-queue for CSR controller:
1. Install `cert-manager`, `cert-manager-csi-driver`, and [example-app](https://github.com/cert-manager/csi-driver/blob/main/deploy/example/example-app.yaml)
2. Add `csi.cert-manager.io/duration: "1m"` to csi `volumeAttribute` in the above `example-app`, to reduce renewal interval. 
    - Check associated `cert-manager-csi-driver` pod logs, you may observe `"Triggering new issuance"` message every 40s (2/3 of duration time).  
3. Now run `kubectl -n cert-manager scale --replicas 0 deployment.apps/cert-manager` to disable CSR controller.
    - Check associated `cert-manager-csi-driver` pod logs, no additional message after `"Triggering new issuance"`. That's because the driver is in an infinite waiting loop.

# Root Cause
In [startRenewalRoutine()](https://github.com/cert-manager/csi-lib/blob/6284a0d2a21d0977bc3298b7f3d429636d63cb8e/manager/manager.go#L597) method, `issue()` will be called upon renewal time, controlled by [RenewalBackoffConfig](https://github.com/cert-manager/csi-lib/blob/6284a0d2a21d0977bc3298b7f3d429636d63cb8e/manager/manager.go#L88-L89)
- each FAILED call will be retried with an interval time set by `RenewalBackoffConfig`: e.g., 10s, 20s, 40s, ..., 
- each PENDING call will wait until `ctx.Timeout` reached

However, in this case, `ctx.Timeout` is undefined as we use `ctx` from this line:
https://github.com/cert-manager/csi-lib/blob/6284a0d2a21d0977bc3298b7f3d429636d63cb8e/manager/manager.go#L608-L609
[...]

PR #46, opened 2023-02-24
Add timeout to renewal issuance logic

7ing:
Add an internal parameter to control issue() call during renewal:
- issueRenewalTimeout: default 60s to align with NodePublishVolume timeout value

This fix prevents inifite loop in a single issue() call when CertificateRequest is in a pending state.

fixes: #45

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -87,11 +87,6 @@ type Options struct {
 
 	// RenewalBackoffConfig configures the exponential backoff applied to certificate renewal failures.
 	RenewalBackoffConfig *wait.Backoff
-	// IssueRenewalTimeout defines timeout value for each issue() call in renewal process
-	IssueRenewalTimeout time.Duration
-	// IssuePollInterval defines an interval time for each subroutine call (check CSR status) within issue() function
-	// Note, the IssuePollInterval should be less than IssueRenewalTimeout
-	IssuePollInterval time.Duration
 }
 
 // NewManager constructs a new manager used to manage volumes containing
@@ -153,16 +148,6 @@ func NewManager(opts Options) (*Manager, error) {
 	if opts.MaxRequestsPerVolume < 0 {
 		return nil, errors.New("MaxRequestsPerVolume cannot be less than zero")
 	}
-	if opts.IssueRenewalTimeout == 0 {
-		opts.IssueRenewalTimeout = 10 * time.Second
-	}
-	if opts.IssuePollInterval == 0 {
-		opts.IssuePollInterval = time.Second
-	}
-	if opts.IssuePollInterval >= opts.IssueRenewalTimeout {
-		return nil, errors.New("IssuePollInterval should be less than IssueRenewalTimeout")
-	}
-	// Check IssuePollInterval is less than IssueRenewalTimeout
 	if len(opts.NodeID) == 0 {
 		return nil, errors.New("NodeID must be set")
 	}
@@ -204,8 +189,7 @@ func NewManager(opts Options) (*Manager, error) {
 		maxRequestsPerVolume: opts.MaxRequestsPerVolume,
 		nodeNameHash:         nodeNameHash,
 		backoffConfig:        *opts.RenewalBackoffConfig,
-		issueRenewalTimeout:  opts.IssueRenewalTimeout,
-		issuePollInterval:    opts.IssuePollInterval,
+		issueRenewalTimeout:  time.Second * 60, // issueRenewalTimeout set to align with NodePublishVolume timeout value
 		requestNameGenerator: func() string {
 			return string(uuid.NewUUID())
 		},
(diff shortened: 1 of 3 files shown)


irbekrm:
Hi @7ing thank you for the PR and sorry for taking so long to review.

Thanks for the good PR description too!
I am, however, not sure about the approach introduced in this PR and cannot see a problem with the current implementation.

>  the driver will wait infinitely instead of exponential backoff retry

What the driver does is repeatedly poll and recheck the status of the CertificateRequest ([here](https://github.com/cert-manager/csi-lib/blob/main/manager/manager.go#L342-L414)). Pending status for a CertificateRequest is an intermediate state, it gets applied, for example in cases where some resources, such as a CA issuer's Secret are not found. When the cluster status changes and cert-manager is able to proceed with issuance, the CertificateRequest will get issued and become ready at which point the csi-driver that has been polling the CertificateRequest will see that it is ready and proceed with provisioning the volume. Same would be true in the scenario described in #45 - once you scale cert-manager replicas back to 1 the CertificateRequest will become ready and the volume will be provisioned. The change suggested by this PR would result in creating additional CertificateRequests every 10 seconds- however this is not necessary, if the original one is still pending then it is due to some issue in the cluster that hasn't been fixed yet and creating additional CertificateRequests that will all remain pending anyway will be unnecessary load for kube apiserver. We should just wait for the original one to become ready as the code already does.

Do let me know if that makes sense?

7ing:
Thank you @irbekrm for reviewing this PR.

> The change suggested by this PR would result in creating additional CertificateRequests every 10 seconds- however this is not necessary, if the original one is still pending then it is due to some issue in the cluster that hasn't been fixed yet and creating additional CertificateRequests that will all remain pending anyway will be unnecessary load for kube apiserver. We should just wait for the original one to become ready as the code already does.

Agree, there is no fundamental issue with original approach other than waiting. Consider the manual approval CR / or long waiting queue, the driver won't print out any other message until it is approved / issued. Adding logs certainly helpful, but probably not inside the loop, as it will flush the log every second. In a long run, need think about how to bubble this message up to the CR object for better visibility. 

This PR needs work with https://github.com/cert-manager/csi-lib/pull/48, so the retry won't create additional CertificateRequests. With that said, if we have `IssueRenewalTimeout`, then we will see error message from:
https://github.com/cert-manager/csi-lib/blob/8ebb8cd1fc7e5e891005188920be0a0aa68e16e0/manager/manager.go#L640-L643

I am okay to close this PR. Just bring this up as an option so we could see exp-back-off log even when CR is in pending status.

irbekrm:
>  Adding logs certainly helpful, but probably not inside the loop, as it will flush the log every second

True, although we could add it so that it is only visible if the verbosity is very high (`log.V(6).Info..`)

> bubble this message up to the CR object for better visibility.

I am not sure if csi-driver _should_ be writing to CR status, in this case I would expect that whoever needs to issue the certificate and considers the CR to be i.e 'pending' to update the status with information why it's 'pending'. The csi-driver does not at this point have any useful information with regards to the CR status. I think we want to communicate to the users that we are waiting for specific CR conditions that weren't found- it feels like logs - or maybe even an event against the pod would be a better suited place for that kind of info?

7ing:
Yup, higher verbose log in the loop is reasonable to me. 

Adding log to pod's event seems better, but requires additional RBAC rule for the driver, which perhaps too much to ask.

irbekrm:
> Adding log to pod's event seems better, but requires additional RBAC rule for the driver, which perhaps too much to ask.

It would require RBAC to create events. I think this should be okay.

Additionally the driver would need to obtain pod's name/namespace. I had a brief look around as to whether anyone does anything similar and found this PR against secrets-store-csi-driver where they attempt to do a similar thing including obtaining pod's reference from the volume ID https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/323 - so perhaps this works well for users.

Let me know what you think, whether events like that would be useful for users. Not sure if that should be included in the csi-lib or maybe should be implementation specific.

7ing:
As a library, probably log message is good enough. I will leave this PR open for a while in case anyone have a different opinion.

7ing:
Rebase code on `v0.5.0`

manager/manager.go
```diff
@@ -87,6 +87,11 @@ type Options struct {
 
 	// RenewalBackoffConfig configures the exponential backoff applied to certificate renewal failures.
 	RenewalBackoffConfig *wait.Backoff
+	// IssueRenewalTimeout defines timeout value for each issue() call in renewal process
+	IssueRenewalTimeout time.Duration
+	// IssuePollInterval defines an interval time for each subroutine call (check CSR status) within issue() function
+	// Note, the issuePollInterval should be less than renewalIssueTimeout
```
munnerz, line 93: Can you add validation into the `NewManager` function to ensure that `IssuePollInterval` is less than `IssueRenewalTimeout`? (so the error happens immediately instead of getting weird/awful spinning edge-cases)

munnerz:
Thanks for this! 😄 

/approve
/lgtm

PR #51, opened 2023-08-31
add pending request cache to allow for resuming in-flight requests that take longer than a single issuance cycle

munnerz:
replaces #48 

Some issuers take a very long time to approve and issue a certificate request. Because the CSI drivers NodePublishVolume call has its own implicit timeout (1 minute) which cannot be changed easily, we would like to be able to 'resume' a request even if it has taken longer than 1 minute to complete.

As a concrete example, say an issuer *always* takes 90s to complete a request (for whatever reason). In these cases, the driver will wait 60s, the context will timeout, 30s later the request will be issued, but upon the NodePublishVolume call being retried we will continuously create a new request.

This is obviously not desirable, so persisting a reference to the `crypto.PrivateKey` in memory allows us to 'resume' the request if it is still usable.

This is different to #48 in that I've avoided modifying large parts of existing code-flows - instead, basically using a map as a cache to lookup an existing private key.

I've also added in an event handler that monitors 'delete' operations on CertificateRequest objects so we can handle the case where another entity deletes requests, so we don't keep persisting stale private keys in memory forever (aka a memory link)

This PR is still WIP, as I need to add a number of tests for it. To do that effectively, I need #46 to be merged so we can timeout the first `issue` call.

cc @7ing @irbekrm

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -496,9 +496,9 @@ func certificateRequestCanBeResumed(req *cmapi.CertificateRequest) bool {
 func (m *Manager) handleRequest(ctx context.Context, volumeID string, meta metadata.Metadata, key crypto.PrivateKey, req *cmapi.CertificateRequest) error {
 	log := m.log.WithValues("volume_id", volumeID)
 
-	// Poll every 1s for the CertificateRequest to be ready
+	// Poll every 200ms for the CertificateRequest to be ready
 	lastFailureReason := ""
-	if err := wait.PollUntilWithContext(ctx, time.Second, func(ctx context.Context) (done bool, err error) {
+	if err := wait.PollUntilWithContext(ctx, time.Millisecond*200, func(ctx context.Context) (done bool, err error) {
 		updatedReq, err := m.lister.CertificateRequests(req.Namespace).Get(req.Name)
 		if apierrors.IsNotFound(err) {
 			// A NotFound error implies something deleted the resource - fail
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -209,6 +209,7 @@ func NewManager(opts Options) (*Manager, error) {
 		reqs, err := lister.List(labels.Everything())
 		if err != nil {
 			janitorLogger.Error(err, "failed listing existing requests")
+			return
 		}
 
 		existsMap := make(map[types.UID]struct{})
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -18,6 +18,7 @@ package manager
 
 import (
 	"context"
+	"crypto"
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
@@ -40,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/clock"
 
 	internalapi "github.com/cert-manager/csi-lib/internal/api"
@@ -157,6 +159,9 @@ func NewManager(opts Options) (*Manager, error) {
 		return nil, fmt.Errorf("building node name label selector: %w", err)
 	}
 
+	// construct the requestToPrivateKeyMap so we can use event handlers below to manage it
+	var requestToPrivateKeyLock sync.Mutex
+	requestToPrivateKeyMap := make(map[types.UID]crypto.PrivateKey)
 	// Create an informer factory
 	informerFactory := cminformers.NewSharedInformerFactoryWithOptions(opts.Client, 0, cminformers.WithTweakListOptions(func(opts *metav1.ListOptions) {
 		opts.LabelSelector = labels.NewSelector().Add(*nodeNameReq).String()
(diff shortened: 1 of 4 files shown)


manager/manager.go
```diff
@@ -339,6 +426,65 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
...
+		return nil, nil
+	}
+
+	// only consider the newest request - we will never resume an older request
+	req := reqs[0]
+	if !certificateRequestCanBeResumed(req) {
+		return nil, nil
+	}
+
+	// TODO: check if this request is still actually valid for the input metadata
```
munnerz, line 464: I've not done this just yet, as I don't think it's quite as important as we may think (as volumeAttributes on a pod are not mutable).

The only case where this could be problematic is if a drivers implementation of `generateRequest` is non-deterministic/can change between calls. To properly handle the wide-range of weird setups users may have, we may actually need to push this comparison function to the driver implementers interface...

manager/manager.go
```diff
@@ -311,10 +377,31 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
 	}
 	log.V(2).Info("Read metadata", "metadata", meta)
 
+	// check if there is already a pending request in-flight for this volume.
+	// if there is, and we still have a copy of its private key in memory, we can resume the existing request and
+	// avoid creating additional CertificateRequest objects.
+	existingReq, err := m.findPendingRequest(meta)
```
munnerz, line 383: This needs adapting to not just return a *pending* request, but also to return a *completed* request if one exists (so that it can be passed to `handleRequest` and actually issued.

manager/manager.go
```diff
@@ -339,6 +426,65 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
...
+	req := reqs[0]
+	if !certificateRequestCanBeResumed(req) {
+		return nil, nil
+	}
+
+	// TODO: check if this request is still actually valid for the input metadata
+	return req, nil
+}
+
+func certificateRequestCanBeResumed(req *cmapi.CertificateRequest) bool {
```
munnerz, line 468: todo: this is inverted 🤦

manager/manager.go
7ing: Consider check privateKey against CSR's public key here ?

munnerz: Given the internal cache is in memory, UIDs are guaranteed to be unique, and the CertificateRequest resource is immutable, I don't think it's actually essential for us to implement the privatekey<>public key check...

I think I am also going to leave this TODO for now, as it isn't really something we handle at all at the moment (and does raise questions around timing, e.g. what if the driver *is* random, when can we ever really stop?). I'd like us to expand on our expectations around how drivers implement generateRequest before over-complicating this code path :)

manager/manager.go
```diff
@@ -157,25 +159,85 @@ func NewManager(opts Options) (*Manager, error) {
...
+	// the request data from its cache.
+	// this routine must be careful to not delete entries from this map that have JUST been added to the map, but haven't
+	// been observed by the lister yet (else it may purge data we want to keep, causing a whole new request cycle).
+	// for now, to avoid this case, we only run the routine every 5 minutes. It would be better if we recorded the time we
+	// added the entry to the map instead, and only purged items from the map that are older that N duration (TBD).
+	janitorLogger := opts.Log.WithName("pending_request_janitor")
+	go wait.Until(func() {
+		reqs, err := lister.List(labels.Everything())
+		if err != nil {
+			janitorLogger.Error(err, "failed listing existing requests")
```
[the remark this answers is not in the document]
munnerz: fixed

manager/manager.go
```diff
@@ -157,25 +159,85 @@ func NewManager(opts Options) (*Manager, error) {
...
+	// create a stop channel that manages all sub goroutines
 	stopCh := make(chan struct{})
+	// begin a background routine which periodically checks to ensure all members of the pending request map actually
+	// have corresponding CertificateRequest objects in the apiserver.
+	// This avoids leaking memory if we don't observe a request being deleted, or we observe it after the lister has purged
+	// the request data from its cache.
+	// this routine must be careful to not delete entries from this map that have JUST been added to the map, but haven't
+	// been observed by the lister yet (else it may purge data we want to keep, causing a whole new request cycle).
+	// for now, to avoid this case, we only run the routine every 5 minutes. It would be better if we recorded the time we
+	// added the entry to the map instead, and only purged items from the map that are older that N duration (TBD).
```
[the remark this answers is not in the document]
munnerz: Yep with a resync period, we *will* definitely see the fact that the CR has been deleted. However, it's not guaranteed that the lister will still have a copy of the object stored - without the object, we can't convert the `namespace/name` to a known UID to look up in the requestToPrivateKeyMap.

Hence, if we don't have access to the UID once we have observed the delete, we won't be able to de-register/remove it from our own internal map.

manager/manager.go
```diff
@@ -259,6 +321,10 @@ type Manager struct {
 	// lister is used as a read-only cache of CertificateRequest resources
 	lister cmlisters.CertificateRequestLister
 
+	// A map that associates a CertificateRequest's UID with its private key.
+	requestToPrivateKeyLock *sync.Mutex
```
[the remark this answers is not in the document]
munnerz: We don't expect to have many concurrent readers so it seemed like a very negligible performance gain (and keeps things a _little_ simpler to use a regular mutex for future readers)

manager/manager.go
```diff
@@ -386,10 +544,12 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
 
 		switch readyCondition.Reason {
 		case cmapi.CertificateRequestReasonIssued:
+			log.V(4).Info("CertificateRequest has been issued!")
 			break
 		case cmapi.CertificateRequestReasonFailed:
 			return false, fmt.Errorf("request %q has failed: %s", updatedReq.Name, readyCondition.Message)
```
7ing, line 550: Do we consider call `m.deletePendingRequestPrivateKey(req.UID)` when the `CertificateRequest` is in `Failed` condition ? It likely to fail again with the same private key. Create a new CR might help resolving the problem if it is due to key reuse issue.

manager/manager.go
```diff
@@ -339,11 +428,78 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
...
+		}
+	}
+	// if there is no Ready condition, the request is still pending processing
+	return true
+}
+
+func (m *Manager) handleRequest(ctx context.Context, volumeID string, meta metadata.Metadata, key crypto.PrivateKey, req *cmapi.CertificateRequest) error {
+	log := m.log.WithValues("volume_id", volumeID)
+
+	// Poll every 200ms for the CertificateRequest to be ready
```
7ing, line 496: Any reason to choose 200ms now? This looks like a typical round tripper time for remote data center query.

munnerz: I had reduced it here as this whole block only ever reads from a local in-memory cache anyway, and it reduced test flakes (as there were a few awkward timing issues where we had timeouts of 2s, but 1s sleeps in between each 'loop' here)

manager/manager.go
munnerz: We'll automatically use a new private key the next time anyway - if a request is failed, it is terminal so will never be returned again by `findPendingRequest` (i.e. it won't be re-used). This will in turn trigger a new CR to be created and new private key generated (or at least, a new call to generatePrivateKey).

The item will be deleted from the map once the CR is deleted, although yep perhaps a future optimisation could be to delete terminal failed items from the map a bit early just to save on memory.. but it shouldn't have any functional difference :)

7ing (approved):

manager/manager.go
```diff
@@ -421,37 +581,55 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
...
+	if err != nil {
+		return err
+	}
+	if len(reqs) <= m.maxRequestsPerVolume {
+		return nil
+	}
+
 	// start at the end of the slice and work back to maxRequestsPerVolume
-	for i := len(reqs) - 1; i >= m.maxRequestsPerVolume-1; i-- {
+	for i := len(reqs) - 1; i > m.maxRequestsPerVolume-1; i-- {
```
JoshVanL, line 632: Why has this been changed? It is because we can now recover the private key between syncs?

test/integration/resume_request_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+func TestResumesExistingRequest_IssuedBetweenPublishCalls(t *testing.T) {
+	testResumesExistingRequest(t, true)
+}
+
+func TestResumesExistingRequest_IssuedDuringPublishCall(t *testing.T) {
+	testResumesExistingRequest(t, false)
+}
```
JoshVanL, line 163: Consider using a type bool enum or passing an options struct so that it is clearer to the reader what `true` and `false` is doing here.

test/integration/resume_request_test.go
```diff
@@ -0,0 +1,183 @@
...
+			return store.WriteMetadata(meta.VolumeID, meta)
+		},
+	})
+	defer stop()
+
+	tmpDir, err := os.MkdirTemp("", "*")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmpDir)
```
JoshVanL, line 75:
```suggestion
	tmpDir := t.TempDir()
```

test/integration/resume_request_test.go
```diff
@@ -0,0 +1,183 @@
...
+			store.WriteFiles(meta, map[string][]byte{
+				"ca":   ca,
+				"cert": chain,
+			})
+			nextIssuanceTime := clock.Now().Add(time.Hour)
+			meta.NextIssuanceTime = &nextIssuanceTime
+			return store.WriteMetadata(meta.VolumeID, meta)
+		},
+	})
+	defer stop()
```
JoshVanL, line 69:
```suggestion
	t.Cleanup(stop)
```

test/integration/resume_request_test.go
```diff
@@ -0,0 +1,183 @@
...
+			"csi.storage.k8s.io/pod.name":      "the-pod-name",
+			"csi.storage.k8s.io/pod.namespace": ns,
+		},
+		TargetPath: tmpDir,
+		Readonly:   true,
+	}
+
+	// create a context that expires in 2s (enough time for at least a single call of `issue`)
+	twoSecondCtx, cancel := context.WithTimeout(ctx, time.Second*2)
+	defer cancel()
```
JoshVanL, line 94: Consider using `t.Cleanup` instead of `defer` for contexts too.
```suggestion
	t.Cleanup(cancel)
```

manager/manager.go
munnerz: yeah exactly - this function was previously doing some ✨ weird ✨ counting logic, which is now fixed (and you can see the behaviour change by taking a look at how the unit tests have changed too)

manager/manager.go
```diff
@@ -157,25 +159,86 @@ func NewManager(opts Options) (*Manager, error) {
...
+			return
+		}
+
+		existsMap := make(map[types.UID]struct{})
+		for _, req := range reqs {
+			existsMap[req.UID] = struct{}{}
+		}
+
+		requestToPrivateKeyLock.Lock()
+		defer requestToPrivateKeyLock.Unlock()
```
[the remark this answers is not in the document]
JoshVanL: This doesn't to be done ^

PR #57, opened 2023-10-26
Upgrade golang to 1.21

7ing:
- Also upgrade all dependencies in go.mod.
- Fix driver/controllerserver.go to work with CSI spec v1.9.0


munnerz:
/hold

this updates the minimum supported version of these dependencies, and prevents this library being built with anything older than go 1.21.

do we want this? Binary dependencies can manage the versions of these libraries themselves, and we (I think?) support these newer versions already if someone chooses to use them in their own ‘require’ directives.

based on MVS, I think we might want to keep all of these at the oldest supported versions rather than the newest to make it easier to import and give those depending on us greater control. What do you think?

7ing:
Just thought would be good to align with `cert-manager` repo dependencies for golang 1.21.

7ing:
rebase from main

PR #61, opened 2024-02-13
Remove unnecessary go expressions

inteon:
Simplifies some expressions (based on IDE warnings).


munnerz (approved):

munnerz:
Thanks! Maybe we should be running lint in CI in future...

/lgtm
/approve

PR #64, opened 2024-03-25
Replace deprecated functions and add missing error checks

inteon:
- replaces deprecated functions & imports
- adds missing error return value checks


munnerz (approved):
/lgtm
/approve

PR #65, opened 2024-03-25
Update atomic writer

inteon:
Update the vendored atomic writer to match upstream.


munnerz:
Awesome, so this also moves the chown step to before the atomic update operation, in turn helping to avoid a race condition where the files may not have the current permission bits set immediately after they are made visible to the workload?

Can you update the title/release note to reflect this behavioural change so it can bubble up into release notes?

munnerz (approved):
/approve
/lgtm
/kind bug

PR #69, opened 2024-11-19
Upgrade dependencies

inteon:
WIP: testing if we can reproduce test failures from https://github.com/cert-manager/csi-driver/pull/322 here.


inteon:
Fixed the found bug in https://github.com/cert-manager/csi-lib/pull/70.

PR #70, opened 2024-11-20
BUGFIX: pass full path to chown

inteon:
Chown was using the `tsDirName` instead of the full path: https://github.com/cert-manager/csi-lib/compare/main...inteon:csi-lib:bugfix?expand=1#diff-ba1745ccd4960836e72c7cf9c559b61ecceab32eb78f3be37d56d8c777c1d141L269

This resulted in `WriteFiles` returning an error. The tests now also check that `WriteFiles` does not return an error.


storage/filesystem.go
```diff
@@ -263,11 +256,18 @@ func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte)
...
+		// Change ownership of the data directory, the timestamp directory and all files within it.
+		paths := []string{
+			f.dataPathForVolumeID(meta.VolumeID),
+			filepath.Join(f.dataPathForVolumeID(meta.VolumeID), tsDirName),
+		}
 		for filename := range files {
-			// Set the uid to -1 which means don't change ownership in Go.
-			if err := os.Chown(filepath.Join(tsDirName, filename), -1, int(*fsGroup)); err != nil {
-				return err
+			paths = append(paths, filepath.Join(f.dataPathForVolumeID(meta.VolumeID), tsDirName, filename))
```
munnerz, line 265: Does it matter that we don't set permission on other files in this directory that we may not have explicitly listed in `files`? Do we _really_ want the properties of 'recursive chown' rather than explicitly chown'ing the directory and any other files we are about to write?

... the only concern I can think of with recursive chown, is if someone does something stupid like `ln -s / /path/to/data/dir/something` as we'd then go and chown everything/could be an 'attack vector' to cause the driver ot spend a lot of time doing something useless/harmful.

The only reason I can think for why we *should* recursive chown all, is to avoid issues when cleaning up... though maybe we are better to have Cleanup basically explicitly delete the files we *do* expect to exist, and then attempt to `rmdir` (which will fail if the directory is non-empty). This is probably a better/safer failure mode, so long as we print the "directory not empty" error message in Cleanup :)

manager/utils_test.go
```diff
@@ -102,10 +103,12 @@ func nothingSignRequest(meta metadata.Metadata, key crypto.PrivateKey, request *
 
 func persistingWriteKeypair(store storage.Interface, clock clock.Clock) WriteKeypairFunc {
 	return func(meta metadata.Metadata, key crypto.PrivateKey, chain []byte, ca []byte) error {
-		store.WriteFiles(meta, map[string][]byte{
+		if err := store.WriteFiles(meta, map[string][]byte{
 			"ca":   ca,
 			"cert": chain,
-		})
+		}); err != nil {
```
munnerz, line 109: Huh.... I wonder why we never checked this error before. I think at some point maybe I wrote WriteFiles to never error? 🤔 

+1 from me, though I think we should explicitly call out in the release notes that we now return errors when writing CA and certificate data in case some implementation relied on this behaviour somehow.

storage/filesystem.go
inteon: The latests solution does not use "recursive chown". Instead we just chown the timestamped dir and the files that we created in that dir.

munnerz: Did we do a recursive chown before?

munnerz:
I wonder what we can do to get signal that this broke csi-driver *sooner*? Perhaps we can add an *optional* test job that fetches the csi-driver and attempts to build it with the PRs version of the csi-lib imported/required? i.e. `git clone csi-driver && go mod replace github.com/cert-manager/csi-lib github.com/cert-manager/csi-lib@v{PR} && make test` or something?

storage/filesystem.go
inteon: No, we did not: https://github.com/cert-manager/csi-lib/pull/65/files

inteon:
> I wonder what we can do to get signal that this broke csi-driver _sooner_? Perhaps we can add an _optional_ test job that fetches the csi-driver and attempts to build it with the PRs version of the csi-lib imported/required? i.e. `git clone csi-driver && go mod replace github.com/cert-manager/csi-lib github.com/cert-manager/csi-lib@v{PR} && make test` or something?

There is a lot we can do to improve this repository, including making the existing tests more real-life.

storage/filesystem.go
```diff
@@ -266,7 +266,7 @@ func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte)
 		// If a fsGroup is defined, Chown all files in the timestamped directory.
 		for filename := range files {
 			// Set the uid to -1 which means don't change ownership in Go.
-			if err := os.Chown(filepath.Join(tsDirName, filename), -1, int(*fsGroup)); err != nil {
+			if err := os.Chown(filepath.Join(f.dataPathForVolumeID(meta.VolumeID), tsDirName, filename), -1, int(*fsGroup)); err != nil {
```
inteon, line 269: This is the minimal fix. This PR taught me about a lot of other possible improvements, they will be part of v0.9.0 instead.
This minimal fix matches what we were doing before: https://github.com/cert-manager/csi-lib/pull/65/files#diff-ba1745ccd4960836e72c7cf9c559b61ecceab32eb78f3be37d56d8c777c1d141L270

SgtCoDFish:
I'm happy to believe this is the minimal fix and it works, but it's not clear to me from this PR why this fix was required by updating the csi-driver deps. What actually changed to cause a break? Can we add that detail to this PR for future reference?

inteon:
> I'm happy to believe this is the minimal fix and it works, but it's not clear to me from this PR why this fix was required by updating the csi-driver deps. What actually changed to cause a break? Can we add that detail to this PR for future reference?

I added the missing link to https://github.com/cert-manager/csi-lib/pull/65, which is where the bug was introduced (+ see the code link in the description for the exact line).

PR #71, opened 2024-11-21
Refactor filesystem.go and adapt tests to use a real file system

inteon:
Adds tests to prevent bugs like https://github.com/cert-manager/csi-lib/pull/70 in the future.

Refactors filesystem.go to only accept absolute paths instead of the weird current relative absolute paths (due to `os.DirFS("/")`).

### DIFF
--- a/storage/filesystem.go
+++ b/storage/filesystem.go
@@ -34,6 +34,14 @@ import (
 	"github.com/cert-manager/csi-lib/third_party/util"
 )
 
+// These variables can be overridden by tests, this makes it easier to test
+// the filesystem implementation as a non-root user.
+var (
+	parentFolderMode     fs.FileMode = 0700
+	volumeFolderMode     fs.FileMode = 0644
+	volumeDataFolderMode fs.FileMode = 0550
+)
+
 const (
 	readWriteUserFileMode        = 0600
 	readOnlyUserAndGroupFileMode = 0440
@@ -42,12 +50,9 @@ const (
 type Filesystem struct {
 	log logr.Logger
 
-	// baseDir is the absolute path to a directory used to store all metadata
+	// basePath is the absolute path to a directory used to store all metadata
 	// about mounted volumes and mount points.
-	baseDir string
-
-	// used by the 'read only' methods
-	fs fs.StatFS
+	basePath string
 
 	// FixedFSGroup is an optional field which will set the gid ownership of all
 	// volume's data directories to this value.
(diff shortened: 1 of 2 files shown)


storage/filesystem.go
```diff
@@ -65,30 +70,44 @@ type Filesystem struct {
...
+	}
+
+	return f, nil
+}
+
+// WARNING: should be used only for testing purposes only, as we don't want to store
+// any sensitive data on disk in production.
+func NewFilesystemOnDisk(log logr.Logger, basePath string) (*Filesystem, error) {
+	if !filepath.IsAbs(basePath) {
+		return nil, fmt.Errorf("baseDir must be an absolute path")
```
inteon, line 100: Starting from this PR, we no longer support the weird relative path notation. All paths have to be absolute.

inteon: see https://github.com/cert-manager/csi-lib/pull/71/files#diff-ebf7d236278a68704d3f90c6e44983c3c663d2beb9ce124a1658cb93ffaa6276R89

storage/filesystem.go
```diff
@@ -239,9 +236,14 @@ func (f *Filesystem) WriteFiles(meta metadata.Metadata, files map[string][]byte)
...
 
+	// Data directory should be read and execute only to the fs user and group.
+	if err := os.MkdirAll(f.dataPathForVolumeID(meta.VolumeID), volumeDataFolderMode); err != nil {
+		return err
+	}
+
 	// If a fsGroup is defined, Chown the directory to that group.
 	if fsGroup != nil {
-		if err := os.Chown(f.dataPathForVolumeID(meta.VolumeID), -1, int(*fsGroup)); err != nil {
+		if err := os.Lchown(f.dataPathForVolumeID(meta.VolumeID), -1, int(*fsGroup)); err != nil {
```
inteon, line 246: Chown vs Lchown:
> If the file is a symbolic link, it changes the uid and gid of the link itself.

This seems more secure. However, since the `dataPathForVolumeID` is a folder, this should not change anything.

storage/filesystem.go
```diff
@@ -171,11 +195,6 @@ func (f *Filesystem) WriteMetadata(volumeID string, meta metadata.Metadata) erro
 func (f *Filesystem) RegisterMetadata(meta metadata.Metadata) (bool, error) {
 	existingMeta, err := f.ReadMetadata(meta.VolumeID)
 	if errors.Is(err, ErrNotFound) {
-		// Ensure directory structure for the volume exists
-		if err := f.ensureVolumeDirectory(meta.VolumeID); err != nil {
```
inteon, line 175: This is now called in `WriteMetadata` instead.

PR #72, opened 2024-11-21
Add fsGroup support

inteon:
Respect the fsGroup option provided as part of the `securityContext.fsGroup`.


driver/controllerserver.go
```diff
@@ -70,15 +70,7 @@ func (cs *controllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacit
...
-		Capabilities: []*csi.ControllerServiceCapability{
-			{
-				Type: &csi.ControllerServiceCapability_Rpc{
-					Rpc: &csi.ControllerServiceCapability_RPC{
-						Type: csi.ControllerServiceCapability_RPC_UNKNOWN,
-					},
-				},
-			},
-		},
+		Capabilities: []*csi.ControllerServiceCapability{},
```
inteon, line 73: Unrelated change: we can skip the UNKNOWN type afaik.

storage/filesystem.go
```diff
@@ -338,8 +338,11 @@ func (f *Filesystem) fsGroupForMetadata(meta metadata.Metadata) (*int64, error)
 		return nil, nil
 	}
 
-	fsGroupStr, ok := meta.VolumeContext[f.FSGroupVolumeAttributeKey]
-	if !ok {
+	fsGroupStr := meta.VolumeContext[f.FSGroupVolumeAttributeKey]
```
munnerz, line 341: Could you add a comment to note we respect the volumeAttribute over the pod configured value? We should probably add a warning message in the log or something if a pod specifies both and they use different values, as this is likely a misconfiguration on the users part.

inteon: I think there might be a valid use case for providing both, the securityContext fsGroup is scoped across all containers and volumes, while the annotation is more granular scoped (just the volume).

inteon: I updated the PR: added test cases and fixed a bug.

wallrj (changes requested):
> Respect the fsGroup option provided as part of the securityContext.fsGroup.

Please explain why this is desirable. What bug will it fix in csi-driver-spiffe? What improvement will it bring to csi-driver-spiffe?
Please show in a csi-driver-spiffe PR how you intend to make use of this.

> Removes the FixedFSGroup field, which does not seem to be used by any implementation.

How do you know that this is not used by any implementation? 
What was the purpose of this field? 
How was it intended to be used?

We should deprecate such fields, right? 
And then remove them in a future release.

inteon:
> > Respect the fsGroup option provided as part of the securityContext.fsGroup.
> 
> Please explain why this is desirable. What bug will it fix in csi-driver-spiffe? What improvement will it bring to csi-driver-spiffe? Please show in a csi-driver-spiffe PR how you intend to make use of this.

When running as a non-root user, you have to make sure the csi-driver mounted files are accessible by the non-root user. This can be done using the `csi.cert-manager.io/fs-group` volume attribute. Kubernetes now has a native `securityContext.fsGroup` option, which we want to support.

> > Removes the FixedFSGroup field, which does not seem to be used by any implementation.
> 
> How do you know that this is not used by any implementation? What was the purpose of this field? How was it intended to be used?
> 
> We should deprecate such fields, right? And then remove them in a future release.

I searched for it on GH and it does not seem to be used.

driver/controllerserver.go
inteon: The unknown value is directly ignored.
In the CSI server implementation, these capabilities are transformed to a map: https://github.com/kubernetes-csi/csi-lib-utils/blob/00c444165f91d6c3ef7b26a98072797a2f11a2c7/rpc/common.go#L84
The map is only checked against using "real" capabilities: https://github.com/kubernetes-csi/external-attacher/blob/4abf4767065b417ebe69dbb5e4636c2a2ed6f2fe/cmd/csi-attacher/main.go#L337

We might want to reconsider advertising the CONTROLLER_SERVICE capability here: https://github.com/cert-manager/csi-lib/blob/449d72f5860ac823dfe0c56e02603119eb232f01/driver/identityserver.go#L68

Issue #60, opened 2024-01-16
Support prometheus metrics


7ing:
Similar to cert-manager controllers (https://cert-manager.io/docs/devops-tips/prometheus-metrics/#scraping-metrics), csi-lib should expose some key metrics for promethues to scrap, like:

- [ ] CertificateRequest status (number issued, approved/denied, failed, etc)
   ```
   certmanager_certificate_ready_status
   ```
- [ ] Near expiry CertificateRequests (how many dates left to renew)
   ```
   certmanager_certificate_expiration_timestamp_seconds
   ```

munnerz:
+1 to this - a few other thoughts/ideas on what we could expose (non-exhaustive, so please add your own ideas!):

1) a metric for "how many pods on this node that reference this `CSI driver` could be useful to understand what the driver 'knows'
2) how many certificates we are 'managing' (which should match to (1), and would be an error/alert if it didn't)
3) expiration time
4) CertificateRequest terminal state reporting

wallrj-cyberark:
Released in https://github.com/cert-manager/csi-lib/releases/tag/v0.10.0. Please test and report back.

PR #73, opened 2024-11-22
Support prometheus metrics

7ing:
certmanager_csi_certificate_request_expiration_timestamp_seconds certmanager_csi_certificate_request_ready_status
certmanager_csi_certificate_request_renewal_timestamp_seconds certmanager_csi_driver_issue_call_count
certmanager_csi_driver_issue_error_count
certmanager_csi_managed_certificate_count
certmanager_csi_managed_volume_count

fixes: #60

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -47,6 +47,7 @@ import (
 	internalapi "github.com/cert-manager/csi-lib/internal/api"
 	internalapiutil "github.com/cert-manager/csi-lib/internal/api/util"
 	"github.com/cert-manager/csi-lib/metadata"
+	"github.com/cert-manager/csi-lib/metrics"
 	"github.com/cert-manager/csi-lib/storage"
 )
 
@@ -89,6 +90,9 @@ type Options struct {
 
 	// RenewalBackoffConfig configures the exponential backoff applied to certificate renewal failures.
 	RenewalBackoffConfig *wait.Backoff
+
+	// Metrics is used for exposing Prometheus metrics
+	Metrics *metrics.Metrics
 }
 
 // NewManager constructs a new manager used to manage volumes containing
@@ -125,6 +129,9 @@ func NewManager(opts Options) (*Manager, error) {
 	}
 	if opts.Log == nil {
 		return nil, errors.New("log must be set")
+	}
+	if opts.Metrics == nil {
+		opts.Metrics = metrics.New(opts.Log)
 	}
 	if opts.MetadataReader == nil {
 		return nil, errors.New("MetadataReader must be set")
@@ -241,6 +248,7 @@ func NewManager(opts Options) (*Manager, error) {
 		metadataReader:          opts.MetadataReader,
 		clock:                   opts.Clock,
 		log:                     *opts.Log,
+		metrics:                 opts.Metrics,
 
 		generatePrivateKey: opts.GeneratePrivateKey,
 		generateRequest:    opts.GenerateRequest,
@@ -375,6 +383,9 @@ type Manager struct {
 	// No thread safety is added around this field, and it MUST NOT be used for any implementation logic.
 	// It should not be used full-stop :).
 	doNotUse_CallOnEachIssue func()
+
+	// metrics is used to expose Prometheus
+	metrics *metrics.Metrics
 }
 
 // issue will step through the entire issuance flow for a volume.
@@ -386,6 +397,9 @@ func (m *Manager) issue(ctx context.Context, volumeID string) error {
 
 	log := m.log.WithValues("volume_id", volumeID)
 	log.Info("Processing issuance")
+
+	// Increase issue count
+	m.metrics.IncrementIssueCallCount(m.nodeNameHash, volumeID)
 
 	if err := m.cleanupStaleRequests(ctx, log, volumeID); err != nil {
 		return fmt.Errorf("cleaning up stale requests: %w", err)
@@ -594,7 +608,7 @@ func (m *Manager) handleRequest(ctx context.Context, volumeID string, meta metad
 	// Calculate the default next issuance time.
 	// The implementation's writeKeypair function may override this value before
 	// writing to the storage layer.
-	renewalPoint, err := calculateNextIssuanceTime(req.Status.Certificate)
+	expiryPoint, renewalPoint, err := getExpiryAndDefaultNextIssuanceTime(req.Status.Certificate)
 	if err != nil {
 		return fmt.Errorf("calculating next issuance time: %w", err)
 	}
@@ -605,6 +619,10 @@ func (m *Manager) handleRequest(ctx context.Context, volumeID string, meta metad
 		return fmt.Errorf("writing keypair: %w", err)
 	}
 	log.V(2).Info("Wrote new keypair to storage")
+
+	// Update the request metrics.
+	// Using meta.NextIssuanceTime instead of renewalPoint here, in case writeKeypair overrides the value.
+	m.metrics.UpdateCertificateRequest(req, expiryPoint, *meta.NextIssuanceTime)
 
 	// We must explicitly delete the private key from the pending requests map so that the existing Completed
 	// request will not be re-used upon renewal.
@@ -656,6 +674,9 @@ func (m *Manager) cleanupStaleRequests(ctx context.Context, log logr.Logger, vol
 				return fmt.Errorf("deleting old certificaterequest: %w", err)
 			}
 		}
+
+		// Remove the CertificateRequest from the metrics.
+		m.metrics.RemoveCertificateRequest(toDelete.Name, toDelete.Namespace)
 
 		log.Info("Deleted CertificateRequest resource", "name", toDelete.Name, "namespace", toDelete.Namespace)
 	}
@@ -756,6 +777,8 @@ func (m *Manager) ManageVolumeImmediate(ctx context.Context, volumeID string) (m
 		// If issuance fails, immediately return without retrying so the caller can decide
 		// how to proceed depending on the context this method was called within.
 		if err := m.issue(ctx, volumeID); err != nil {
+			// Increase issue error count
+			m.metrics.IncrementIssueErrorCount(m.nodeNameHash, volumeID)
 			return true, err
 		}
 	}
@@ -783,6 +806,8 @@ func (m *Manager) manageVolumeIfNotManaged(volumeID string) (managed bool) {
 	// construct a new channel used to stop management of the volume
 	stopCh := make(chan struct{})
 	m.managedVolumes[volumeID] = stopCh
+	// Increase managed volume count for this driver
+	m.metrics.IncrementManagedVolumeCount(m.nodeNameHash)
 
 	return true
 }
@@ -799,6 +824,10 @@ func (m *Manager) startRenewalRoutine(volumeID string) (started bool) {
 		log.Info("Volume not registered for management, cannot start renewal routine...")
 		return false
 	}
+
+	// Increase managed certificate count for this driver.
+	// We assume each volume will have one certificate to be managed.
+	m.metrics.IncrementManagedCertificateCount(m.nodeNameHash)
 
 	// Create a context that will be cancelled when the stopCh is closed
 	ctx, cancel := context.WithCancel(context.Background())
@@ -835,6 +864,8 @@ func (m *Manager) startRenewalRoutine(volumeID string) (started bool) {
 						defer issueCancel()
 						if err := m.issue(issueCtx, volumeID); err != nil {
 							log.Error(err, "Failed to issue certificate, retrying after applying exponential backoff")
+							// Increase issue error count
+							m.metrics.IncrementIssueErrorCount(m.nodeNameHash, volumeID)
 							return false, nil
 						}
 						return true, nil
@@ -874,6 +905,14 @@ func (m *Manager) UnmanageVolume(volumeID string) {
 	if stopCh, ok := m.managedVolumes[volumeID]; ok {
 		close(stopCh)
 		delete(m.managedVolumes, volumeID)
+		if reqs, err := m.listAllRequestsForVolume(volumeID); err == nil {
+			// Remove the CertificateRequest from the metrics with the best effort.
+			for _, req := range reqs {
+				if req != nil {
+					m.metrics.RemoveCertificateRequest(req.Name, req.Namespace)
+				}
+			}
+		}
 	}
 }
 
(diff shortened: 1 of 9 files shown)


metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+		driverIssueErrorCount:                driverIssueErrorCount,
+		managedVolumeCount:                   managedVolumeCount,
+		managedCertificateCount:              managedCertificateCount,
+	}
+
+	return m
+}
+
+// NewServer registers Prometheus metrics and returns a new Prometheus metrics HTTP server.
+func (m *Metrics) NewServer(ln net.Listener) *http.Server {
```
munnerz, line 151: This isn't called outside of test cases, and I guess that is by design as it is expected that the corresponding implementation should call `NewServer` on `metrics.Metrics` to register their Listener.

Could you possibly update the `example/` implementation in the root of this repository to demonstrate how to actually add the `/metrics` endpoint? I also wonder if the Managed should be extended to be able to auto-serve this endpoint in cases where a user doesn't _need_ to provide their own listener (but _does_ want metrics to be served).

metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+				Subsystem: subsystem,
+				Name:      "managed_certificate_count",
+				Help:      "The number of certificates managed by the csi driver.",
+			},
+			[]string{"node"},
+		)
+	)
+
+	// Create Registry and register the recommended collectors
+	registry := prometheus.NewRegistry()
```
munnerz, line 128: Could there be cases where a user wants to provide their own Registry and have these metrics registered into it, so they can be served alongside driver-specific metrics? This would make the library more composable with existing drivers that may serve their own metrics already.

metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	registry.MustRegister(
+		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
+		collectors.NewGoCollector(),
```
munnerz, line 131: I feel like these *should* be auto-registered, but I think the `Metrics` struct itself should be able to be setup so that it *doesn't* embed these Go metrics into them (i.e. so it only adds in the csi-lib specific metrics to the registry).

Perhaps we need some 'basic' constructor function for the csi-lib metrics, with some form of `NewRecommendedMetrics` function that calls the basic one, but also adds in standard Go-esque metrics? Not sure :)

metrics/metrics.go
munnerz: What if we renamed this to `Register(*prometheus.Registry)` rather than tying the `net.Listener` logic into the registration?

We can always have/find some kind of `csihelpers.Handle(*http.Server, *prometheus.Registry)` function elsewhere then, which needn't be opinionated about csi-lib.

munnerz (commented):
This looks great, thanks Jing 🙌 the integration and unit tests here make it far easier to review confidently!

My main questions/concerns are around the construction logic in the `metrics` subpackage, which I think we need to decouple from `net.Listener` (and allow more flexibility for projects that already have their own `prometheus.Registry` they'd like to re-use).

metrics/metrics.go
7ing: > Could you possibly update the example/ implementation in the root of this repository to demonstrate how to actually add the /metrics endpoint? 

Yes, it is there. And modified accordingly based on recent changes. 
https://github.com/7ing/csi-lib/blob/abf15631238fa809d10d9c206178c414d9495b4f/test/integration/metrics_test.go#L83-L95

7ing: Thanks for the suggestions. I have changed to use a `DefaultHandler` instead, which could be served as a reference implementation for http handler.

metrics/metrics.go
7ing: Good idea, code changed per your suggestion.

metrics/metrics.go
7ing: Agree, I will remove those default registry and let the caller to pass their Registry instead.

7ing:
@munnerz Thank you for your valuable inputs. Sorry took so long to make the change. But I guess this version addressed most of your concerns.

test/integration/metrics_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Get the CSR name
+	req, err := testutil.WaitAndGetOneCertificateRequestInNamespace(ctx, opts.Client, testNamespace)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Should expose that CertificateRequest as ready with expiry and renewal time
+	// node="f56fd9f8b" is the hash value of "test-node" defined in driver_testing.go
+	expectedOutputTemplate := `# HELP certmanager_csi_certificate_request_expiration_timestamp_seconds The date after which the certificate request expires. Expressed as a Unix Epoch Time.
```
inteon, line 215: Will `certmanager_csi_...` be the prefix for the metrics when exposed in https://github.com/cert-manager/csi-driver?
Can we make sure they are all using similar names? Does this happen automatically?
Could you provide an example of before and after for https://github.com/cert-manager/csi-driver with these changes applied?

7ing: Thank you @inteon for your review.

Yes, `certmanager_csi_...` will be the prefix for the metrics when in the csi-driver. It is part of the definition from:
https://github.com/7ing/csi-lib/blob/b9186bad5b6f9af9bc93dbd28571d2d9219700e6/metrics/metrics.go#L27-L31

We choose this name based on cert-manager.io definition: https://github.com/cert-manager/cert-manager/blob/5e09ef6c0552df0bde64746c735cb1ff324b6261/pkg/metrics/metrics.go#L44
All cert-manager controllers have `certmanager_..` prefix.

Our https://github.com/cert-manager/csi-driver does not have any certmanager related metrics. Currently it only serves k8s components metrics, like cpu / mem etc. That's why this PR exist.

erikgb:
@7ing We have made some major dependency upgrades in this module. Are you able to rebase your PR, preparing for another round of review? Sorry for the inconvenience and for the delays in review. 😒

7ing:
@erikgb done with rebase

metrics/certificaterequest.go
```diff
@@ -0,0 +1,102 @@
...
+		}).Set(value)
+	}
+}
+
+// RemoveCertificateRequest will delete the CertificateRequest metrics from continuing to be exposed.
+func (m *Metrics) RemoveCertificateRequest(name, namespace string) {
+	m.certificateRequestExpiryTimeSeconds.DeletePartialMatch(prometheus.Labels{"name": name, "namespace": namespace})
+	m.certificateRequestRenewalTimeSeconds.DeletePartialMatch(prometheus.Labels{"name": name, "namespace": namespace})
+	m.certificateRequestReadyStatus.DeletePartialMatch(prometheus.Labels{"name": name, "namespace": namespace})
+}
```
erikgb, line 102: I think we should use a Prometheus Collector for metrics derived from API resources. This will prevent bugs like the one fixed by https://github.com/cert-manager/cert-manager/pull/7856. API resources should be cached in all controllers, so just querying the API, which should hit cache, is not a problem.

erikgb (commented):
This is great stuff! Thanks @7ing! I did my first pass on this PR now, and I think I would prefer using a Prometheus collector to avoid the add/remove metrics for metrics based on API resources. Please take a look and let me know what you think! It's not a blocker, but a pattern we are adopting in cert-manager projects nowadays.

metrics/certificaterequest.go
hjoshi123: I kind of agree with erik on this one.. we should migrate all of our metrics related to API to collectors.. this makes it easier to also cleanup the metrics when the resources are gone

metrics/metrics.go
```diff
@@ -0,0 +1,111 @@
...
+	log      logr.Logger
+	registry *prometheus.Registry
+
+	driverIssueCallCountTotal   *prometheus.CounterVec
+	driverIssueErrorCountTotal  *prometheus.CounterVec
+	certificateRequestCollector prometheus.Collector
+}
+
+// New creates a Metrics struct and populates it with prometheus metric types.
+func New(logger *logr.Logger, registry *prometheus.Registry) *Metrics {
```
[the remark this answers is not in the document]
7ing: In the example, I create the registry and start the metrics server in the same function. It can be separated and provided by the user as input as well. Really depends on how the driver implementation specifies. 
@erikgb @hjoshi123 hope this version clarifies your questions.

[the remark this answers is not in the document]
7ing: @hjoshi123 I have removed the `SetupCertificateRequestCollector` function. So user need to call New(*) func to initialize all metrics data.

7ing:
@erikgb @hjoshi123 I believe this one close to the final version. Let me know if you have further questions. Thanks.

maelvls:
Hi! I'll ask if someone from CyberArk can review this PR.

manager/manager.go
```diff
@@ -375,6 +380,9 @@ type Manager struct {
 	// No thread safety is added around this field, and it MUST NOT be used for any implementation logic.
 	// It should not be used full-stop :).
 	doNotUse_CallOnEachIssue func()
```
erikgb: Can/should this field be removed now, or is this a follow-up? Ref. https://github.com/cert-manager/csi-lib/blob/f3a2df0cf6a71e04bd9a5467fb89dab2e58e5a1a/manager/manager.go#L382-L385

metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	internalapiutil "github.com/cert-manager/csi-lib/internal/api/util"
+	"github.com/cert-manager/csi-lib/storage"
+)
+
+const (
+	// Namespace is the namespace for csi-lib metric names
+	namespace = "certmanager"
+	subsystem = "csi"
```
erikgb: Could it make sense to make the subsystem configurable from the application using csi-lib, with a good default? I am just thinking how this might appear if two CSI drivers using this lib are running in the same environment. It probably doesn't make much sense, but if two drivers using this lib are producing the same metrics, I suggest dropping the `driver_` prefix from the metrics name (commented below).

metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+			[]string{"node", "volume"},
+		)
+
+		// driverIssueErrorCountTotal is a Prometheus counter for the number of errors encountered
+		// during the driver issue() calls.
+		driverIssueErrorCountTotal = prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      "driver_issue_error_count_total",
```
erikgb:
```suggestion
				Name:      "issue_errors_total",
```

If we want to keep `driver_` in the full metrics name, I suggest including it in the subsystem part above.

Naming is hard, but I tried to read https://prometheus.io/docs/practices/naming/, and I think count/total is almost the same.

metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	mux.Handle("/metrics", promhttp.HandlerFor(m.registry, promhttp.HandlerOpts{}))
+
+	return mux
+}
+
+func (m *Metrics) SetupCertificateRequestCollector(nodeId string, metadataReader storage.MetadataReader, certificateRequestLister cmlisters.CertificateRequestLister) {
+	m.certificateRequestCollector = NewCertificateRequestCollector(internalapiutil.HashIdentifier(nodeId), metadataReader, certificateRequestLister)
+	m.registry.MustRegister(m.certificateRequestCollector)
+}
+
```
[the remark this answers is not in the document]
erikgb: I agree with Copilot here. This function appears superfluous after the latest changes.

manager/manager.go
7ing: This should be a follow-up question in a different issue / PR.

metrics/metrics.go
7ing: Don't wanna make things complicated. The metrics data can be separated by namespace / node label in most cases.

examples/simple/main.go
```diff
@@ -89,7 +94,7 @@ func main() {
 		panic("-data-root must be set")
 	}
 
-	log := klogr.New()
+	log := klog.Background()
```
erikgb:
```suggestion
	log := klog.TODO()
```

Nit, but according to the docs this is "more correct".

metrics/metrics.go
```diff
@@ -0,0 +1,119 @@
...
+		issueErrorsTotal = prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      "issue_errors_total",
+				Help:      "The number of errors encountered during the driver issue() calls.",
+			},
+			[]string{"node", "volume"},
+		)
+	)
```
erikgb:
```suggestion
		// issueRequestsTotal is a Prometheus counter for the number of issue() calls made by the driver.
		issueRequestsTotal := prometheus.NewCounterVec(
			prometheus.CounterOpts{
				Namespace: namespace,
				Subsystem: subsystem,
				Name:      "issue_requests_total",
				Help:      "The number of issue() calls made by the driver.",
			},
			[]string{"node", "volume"},
		)

		// issueErrorsTotal is a Prometheus counter for the number of errors encountered
		// during the driver issue() calls.
		issueErrorsTotal := prometheus.NewCounterVec(
			prometheus.CounterOpts{
				Namespace: namespace,
				Subsystem: subsystem,
				Name:      "issue_errors_total",
				Help:      "The number of errors encountered during the driver issue() calls.",
			},
			[]string{"node", "volume"},
		)
```

Nit: Should make the code easier to read with less nesting.

erikgb (commented):
/lgtm
/cc @wallrj 

The code looks good to me. @wallrj has volunteered to test this in csi-driver through https://github.com/cert-manager/csi-driver/pull/502.

metrics/certificaterequest_collector.go
```diff
@@ -0,0 +1,249 @@
...
+	certRequestExpirationTimestampSeconds = prometheus.NewDesc("certmanager_csi_certificate_request_expiration_timestamp_seconds", "The timestamp after which the certificate request expires, expressed in Unix Epoch Time.", []string{"name", "namespace", "issuer_name", "issuer_kind", "issuer_group"}, nil)
+	certRequestRenewalTimestampSeconds    = prometheus.NewDesc("certmanager_csi_certificate_request_renewal_timestamp_seconds", "The timestamp after which the certificate request should be renewed, expressed in Unix Epoch Time.", []string{"name", "namespace", "issuer_name", "issuer_kind", "issuer_group"}, nil)
+	managedVolumeCountTotal               = prometheus.NewDesc("certmanager_csi_managed_volume_count_total", "The total number of managed volumes by the csi driver.", []string{"node"}, nil)
+	managedCertRequestCountTotal          = prometheus.NewDesc("certmanager_csi_managed_certificate_request_count_total", "The total number of managed certificate requests by the csi driver.", []string{"node"}, nil)
```
wallrj-cyberark: I deployed the driver, built from https://github.com/cert-manager/csi-driver/pull/502 and observed these new metrics, before creating any Pods to use the driver:


```
$ kubectl get --raw "/api/v1/namespaces/cert-manager/pods/${POD_NAME}:9402/proxy/metrics"  | grep certmanager
# HELP certmanager_csi_managed_certificate_request_count_total The total number of managed certificate requests by the csi driver.
# TYPE certmanager_csi_managed_certificate_request_count_total counter
certmanager_csi_managed_certificate_request_count_total{node="54bb9576d6"} 0
# HELP certmanager_csi_managed_volume_count_total The total number of managed volumes by the csi driver.
# TYPE certmanager_csi_managed_volume_count_total counter
certmanager_csi_managed_volume_count_total{node="54bb9576d6"} 0
```

metrics/certificaterequest_collector.go
```diff
@@ -0,0 +1,249 @@
...
+			return nil, err
+		}
+		if volumeMetadata.NextIssuanceTime != nil {
+			nextIssuanceTimeMap[internalapiutil.HashIdentifier(id)] = *volumeMetadata.NextIssuanceTime
+		}
+	}
+	return nextIssuanceTimeMap, nil
+}
+
+func (cc *CertificateRequestCollector) updateManagedVolumeCount(count int, ch chan<- prometheus.Metric) {
```
hjoshi123: @7ing I feel this should be `Gauge` instead of a `Counter` since we calculate `len(nextIssuanceTimeMap)` each time. I would love to know why it should be a Counter.

hjoshi123:
The code looks good to me now. Thank you @7ing for making the changes!!
/lgtm

/cc @erikgb @wallrj-cyberark

PR #91, opened 2025-08-13
Manual "self upgrade"

erikgb:
This is a manual alternative to https://github.com/cert-manager/csi-lib/pull/90, as the self-upgrade workflow seems to always be failing now. Probably because of workflow updates, but this works in other projects..... https://github.com/cert-manager/csi-lib/actions/workflows/make-self-upgrade.yaml


driver/driver.go
```diff
@@ -64,9 +65,9 @@ type Options struct {
 	ContinueOnNotReady bool
 }
 
-func New(endpoint string, log logr.Logger, opts Options) (*Driver, error) {
+func New(ctx context.Context, endpoint string, log logr.Logger, opts Options) (*Driver, error) {
```
wallrj-cyberark: Breaking change. I've added this to the release notes. 
 * https://github.com/cert-manager/csi-lib/releases/tag/v0.10.0

Issue #234, opened 2026-06-09
Proposal: distinguish gate-pending from issuance-error in the renewal backoff loop


patyogesh20:
## Context
`ReadyToRequestFunc` (introduced in PR #20) lets a driver defer issuance until pod-side conditions are met. cert-manager/csi-driver PR [#638](https://github.com/cert-manager/csi-driver/pull/638) recently added a `--pod-readiness-gate` flag that exposes this for IP-SAN use cases (the kubelet/CNI race with NodePublishVolume).

  Now that there's a real-world consumer of `ReadyToRequestFunc`, the question of how csi-lib treats "not yet ready" vs. "issuance failed" has surfaced concretely. I'd like maintainer direction on the right fix before opening a PR.

  ## Problem

  In `manager.issue()`, a `false` return from `ReadyToRequestFunc` is converted to an [error](https://github.com/cert-manager/csi-lib/blob/a3e4b56d5dde8db5da310ffc7730cdaa5487703a/manager/manager.go#L424). This error flows back to `startRenewalRoutine`, where it's wrapped by `wait.ExponentialBackoffWithContext(ctx, m.backoffConfig, ...)`. `m.backoffConfig` is the renewal backoff — defaults: 30s base, 5min cap, factor 2.0. Those defaults are appropriate for issuance failures (signer down, CSR rejected, apiserver blip etc.) 

  But they are **wrong** for the gate-pending state, where the underlying condition (pod IP from CNI, condition flipped, annotation written) typically resolves in seconds. Because the same backoff applies, pods wait 30s+ for the next gate evaluation when the gate would have passed in 2s.

  ## Evidence (production logs)

  From a production cluster running csi-driver with `--pod-readiness-gate=pod-ip:ipv6`:

  01:05:16.195  NodePublishVolume RPC received
  01:05:17.199  Triggering new issuance
  01:05:17.203  Failed: "pod has no ipv6 address yet" → backoff
  01:05:59.922  Triggering new issuance       (+42.7s wasted in backoff)
  01:06:00.408  CertificateRequest created
  01:06:00.419  Keypair written

  Wall-clock from `NodePublishVolume` to cert available: 44s. Actual issuance work: ~1.5s. The other 42.7s is `RenewalBackoff` sleeping on a condition (pod IPv6) that resolved within ~2s at the OS level.
  Tail-case from the same log stream: one pod waited 4m 43s across three backoff cycles before its IPv6 appeared and issuance succeeded.

[...]

PR #235, opened 2026-06-10
manager: separate backoff for gate-pending vs issuance failures

patyogesh20:
Implements #234

When `ReadyToRequestFunc` returns false, the renewal loop was synthesising an error and routing it through the same exponential backoff used for genuine issuance failures (30s base, 5min cap). That cadence is correct for protecting a sick signer from a thundering herd, but wrong for a wait state where the underlying condition (CNI assigning a pod IP, a controller flipping a pod condition) typically resolves in seconds.

  This PR:
  - Adds an exported `ErrNotReadyToRequest` sentinel.
  - Adds `GateBackoffConfig` to `manager.Options` (nil defaults to 1s base, factor 2.0, jitter 0.5, cap 10s).
  - Wraps the readyToRequest false return inside `issue()` with the sentinel.
  - Extracts the renewal loop's retry into `attemptIssuanceIfDue`, which routes by error class via `errors.Is` and resets each backoff when the class changes so they decay independently.

## Tests
  Added unit tests
  All existing tests pass (no regression)

## Backwards compatibility
  Consumers using `AlwaysReadyToRequest` (the default) never enter the gate-pending path, so behaviour is identical. Consumers with a custom `ReadyToRequestFunc` (e.g. cert-manager/csi-driver post-#638) inherit the new default gate backoff automatically, which is the intended fix.

### DIFF
--- a/manager/interfaces.go
+++ b/manager/interfaces.go
@@ -29,13 +29,16 @@ import (
 	"github.com/cert-manager/csi-lib/metadata"
 )
 
-// ErrNotReadyToRequest is returned by the renewal loop's issue() when
-// ReadyToRequestFunc reports the driver is not yet ready (a wait state, not a
-// failure). The renewal loop uses errors.Is to distinguish this from real
-// issuance errors (signer down, apiserver error, denial) and applies a
-// separate, faster backoff appropriate for waiting on pod state. Consumers
-// can also use errors.Is to surface this case in their own observability.
-var ErrNotReadyToRequest = errors.New("driver is not ready to request a certificate")
+// errNotReadyToRequest is returned by issue() when ReadyToRequestFunc reports
+// the driver is not yet ready (a wait state, not a failure). The renewal loop
+// uses errors.Is to distinguish this from real issuance errors (signer down,
+// apiserver error, denial) and applies a separate, faster backoff appropriate
+// for waiting on pod state.
+//
+// Kept unexported because issue() is unexported and no current consumer needs
+// to detect this case via errors.Is. Can be promoted to exported if a real
+// use case appears.
+var errNotReadyToRequest = errors.New("driver is not ready to request a certificate")
 
 // GeneratePrivateKeyFunc returns a private key to be used for issuance of the
 // given request.
--- a/manager/interfaces.go
+++ b/manager/interfaces.go
@@ -19,6 +19,7 @@ package manager
 import (
 	"crypto"
 	"crypto/x509"
+	"errors"
 	"time"
 
 	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
@@ -27,6 +28,14 @@ import (
 
 	"github.com/cert-manager/csi-lib/metadata"
 )
+
+// ErrNotReadyToRequest is returned by the renewal loop's issue() when
+// ReadyToRequestFunc reports the driver is not yet ready (a wait state, not a
+// failure). The renewal loop uses errors.Is to distinguish this from real
+// issuance errors (signer down, apiserver error, denial) and applies a
+// separate, faster backoff appropriate for waiting on pod state. Consumers
+// can also use errors.Is to surface this case in their own observability.
+var ErrNotReadyToRequest = errors.New("driver is not ready to request a certificate")
 
 // GeneratePrivateKeyFunc returns a private key to be used for issuance of the
 // given request.
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -96,8 +96,10 @@ type Options struct {
 	// assign a pod IP), not a failure, and warrants a faster retry cadence than
 	// genuine issuance errors. The renewal loop selects between this backoff
 	// and RenewalBackoffConfig based on whether the issue() error wraps
-	// ErrNotReadyToRequest.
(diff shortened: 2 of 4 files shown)


manager/interfaces.go
```diff
@@ -28,6 +29,14 @@ import (
 	"github.com/cert-manager/csi-lib/metadata"
 )
 
+// ErrNotReadyToRequest is returned by the renewal loop's issue() when
+// ReadyToRequestFunc reports the driver is not yet ready (a wait state, not a
+// failure). The renewal loop uses errors.Is to distinguish this from real
+// issuance errors (signer down, apiserver error, denial) and applies a
+// separate, faster backoff appropriate for waiting on pod state. Consumers
+// can also use errors.Is to surface this case in their own observability.
+var ErrNotReadyToRequest = errors.New("driver is not ready to request a certificate")
```
SgtCoDFish: suggestion: I'm not sure if it lines up to have this exported publicly but to have `issue()` be private. Since users don't need to directly handle errors from `issue()`, would it make sense to change `ErrNotReadyToRequest` to `errNotReadyToRequest`? It's easy to expose it later if we need to.

manager/manager.go
```diff
@@ -821,45 +848,69 @@ func (m *Manager) startRenewalRoutine(volumeID string) (started bool) {
...
+	if meta.NextIssuanceTime != nil && !m.clock.Now().After(*meta.NextIssuanceTime) {
+		return
+	}
+
+	issuanceBackoff := m.backoffConfig
+	gateBackoff := m.gateBackoffConfig
+	for {
+		if ctx.Err() != nil {
+			return
+		}
```
SgtCoDFish: suggestion: Isn't this check of `ctx.Err` redundant? We handle the context in the select statement at the bottom of the loop

manager/manager.go
```diff
@@ -91,6 +91,15 @@ type Options struct {
 	// RenewalBackoffConfig configures the exponential backoff applied to certificate renewal failures.
 	RenewalBackoffConfig *wait.Backoff
 
+	// GateBackoffConfig configures the backoff applied when ReadyToRequestFunc
+	// returns false. This is an expected wait state (e.g. waiting for CNI to
+	// assign a pod IP), not a failure, and warrants a faster retry cadence than
+	// genuine issuance errors. The renewal loop selects between this backoff
+	// and RenewalBackoffConfig based on whether the issue() error wraps
+	// ErrNotReadyToRequest.
+	// If nil, defaults to: 1s base, factor 2.0, jitter 0.5, cap 10s.
```
SgtCoDFish: suggestion: I think it's probably important to say that `steps` is effectively infinite by default? If the user looks at the docs for `wait.Backoff` they could wonder what we set that value to.

manager/manager.go
```diff
@@ -91,6 +91,15 @@ type Options struct {
...
+	GateBackoffConfig *wait.Backoff
```
SgtCoDFish, line 103: note(non-blocking): It was maybe a mistake for `RenewalBackoffConfig` to use the type `*wait.Backoff` because it means the values present in our config depend on an external package, and it makes it awkward for us to handle changes in the `wait` package.

I'd have preferred us to copy the backoff config struct and then construct a `wait.Backoff` from our own struct.

But obviously, we already depend on that type for `RenewalBackoffConfig` so it's fine to use it here for `GateBackoffConfig`. I'm just noting my thoughts here.

SgtCoDFish (changes requested):
Few bits but they're small, this looks good!

manager/interfaces.go
patyogesh20: Agreed, we can export it later if required. Unexported in the latest commit.

manager/manager.go
patyogesh20: You're right, the bottom select handles this. Removed.

manager/manager.go
patyogesh20: Good catch. Being explicit about it is important here. Added a note matching the style of the existing `RenewalBackoffConfig` comment.

manager/manager.go
patyogesh20: I see the point regarding option values dependent on this upstream struct. However, copied struct would be the actual anti-pattern IMO. 

csi-lib's `Options` already exposes plenty of external types e.g. [clock.Clock](https://github.com/cert-manager/csi-lib/blob/3222aad83d3f711ebefadd6eb183d1e21f39cade/manager/manager.go#L70), [logr.Logger](https://github.com/cert-manager/csi-lib/blob/3222aad83d3f711ebefadd6eb183d1e21f39cade/manager/manager.go#L73). If we apply the same policy/style "_no library types in public config_", applying it consistently would be a much bigger refactor than this.

In fact, I would argue in favour of "what if `wait.Backoff` changes" concern. If the struct adds a new field, passthrough means we inherit the capability automatically. With a copied type we have to mirror it after the fact, and the lag between mirror and reality is itself a bug. Besides, we still depend on `wait.Backoff` internally when we construct one to call `Step()`. 

Happy to revisit if there's a specific `wait.Backoff` change OR relying on external package & exposing them via our config has caused pain.

SgtCoDFish: No action needed - I'm just thinking out loud !

SgtCoDFish (approved):
/lgtm
/approve

I don't see a reason to hold this up. Thank you for this!

PR #246, opened 2026-06-23
fix: redact ServiceAccount bearer tokens from log output

FelixPhipps:
# Description:

protosanitizer.StripSecrets only redacts protobuf fields annotated csi_secret. The volume_context field — where kubelet injects the mounting pod's ServiceAccount bearer token when spec.tokenRequests is set — carries no such annotation and passed through untouched. This caused live SA tokens to appear in logs at default verbosity on any NodePublishVolume handler error (readOnly unset, issuer timeout, etc.), and at -v=2/-v=3 on every issuance and renewal.

Changes:
  - driver/server.go — redactSATokens() proto-clones the request and replaces the token value with [REDACTED] before passing to protosanitizer.StripSecrets. The handler receives the original unmodified request.
  - metadata/metadata.go — implements logr.Marshaler on Metadata so any logr backend automatically redacts the token when logging the struct (fixes the manager.go:398 sink). Exports SATokenVolumeContextKey as the canonical constant.
  - manager/util/tokenrequest.go — uses the new constant instead of the duplicated string literal.
  - driver/server_test.go & metadata/metadata_test.go — Automated testing added.

  Manual test output using klog (the production logging backend) against all three sinks:

  # Sink #3 — error level, default verbosity (fires in production on any handler failure)
  E server.go:113] "failed processing request" err="pod.spec.volumes[].csi.readOnly must be set to 'true'"
    rpc_method="/csi.v1.Node/NodePublishVolume"
    request={"volume_context":{"csi.storage.k8s.io/serviceAccount.tokens":"[REDACTED]",...},...}

  # Sink #1 — V(3) entry log
  I server.go:110] "handling request"
    request={"volume_context":{"csi.storage.k8s.io/serviceAccount.tokens":"[REDACTED]",...},...}

  # Sink #2 — V(2) Read metadata (manager.go:398)
  I] "Read metadata"
    metadata={"volumeContext":{"csi.storage.k8s.io/serviceAccount.tokens":"[REDACTED]",...}}

  Tests also confirm the handler receives the original unredacted token so ClientForMetadataTokenRequestEmptyAud continues to work.

### DIFF
--- a/driver/server.go
+++ b/driver/server.go
@@ -23,10 +23,12 @@ import (
 	"os"
 	"strings"
 
+	"github.com/cert-manager/csi-lib/metadata"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/go-logr/logr"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"google.golang.org/grpc"
+	protopb "google.golang.org/protobuf/proto"
 )
 
 type GRPCServer struct {
@@ -102,7 +104,8 @@ func parseEndpoint(ep string) (string, string, error) {
 
 func loggingInterceptor(log logr.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
-		log := log.WithValues("rpc_method", info.FullMethod, "request", protosanitizer.StripSecrets(req))
+		sanitized := protosanitizer.StripSecrets(redactSATokens(req))
+		log := log.WithValues("rpc_method", info.FullMethod, "request", sanitized)
 		log.V(3).Info("handling request")
 		resp, err := handler(ctx, req)
 		if err != nil {
@@ -112,4 +115,20 @@ func loggingInterceptor(log logr.Logger) grpc.UnaryServerInterceptor {
 		}
 		return resp, err
 	}
+}
+
+// redactSATokens returns a copy of req with the ServiceAccount bearer token
+// redacted (replaced with "[REDACTED]") in the VolumeContext of a
+// NodePublishVolumeRequest. All other request types are returned as-is. The original request is never mutated.
+func redactSATokens(req any) any {
+	npvr, ok := req.(*csi.NodePublishVolumeRequest)
+	if !ok {
+		return req
+	}
+	if _, hasToken := npvr.GetVolumeContext()[metadata.SATokenVolumeContextKey]; !hasToken {
+		return req
+	}
+	sanitized := protopb.Clone(npvr).(*csi.NodePublishVolumeRequest)
+	sanitized.VolumeContext[metadata.SATokenVolumeContextKey] = "[REDACTED]"
+	return sanitized
 }
--- a/driver/server_test.go
+++ b/driver/server_test.go
@@ -1,0 +1,209 @@
+/*
+Copyright 2026 The cert-manager Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc"
+	"k8s.io/klog/v2/ktesting"
+
+	"github.com/cert-manager/csi-lib/metadata"
+)
+
+// testToken is a realistic kubelet-injected serviceAccount.tokens blob.
+// "secret" is a distinctive substring used throughout to detect leaks.
+const testToken = `{"":{"token":"eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.secret","expirationTimestamp":"2099-01-01T00:00:00Z"}}`
+
+func requireNoTokenInOutput(t *testing.T, output string) {
+	t.Helper()
+	for _, needle := range []string{"secret", "eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9"} {
+		if strings.Contains(output, needle) {
+			t.Errorf("SA token leaked into log output (found %q):\n%s", needle, output)
+		}
+	}
+}
+
+// TestRedactSATokens_NodePublishVolumeRequest verifies that redactSATokens
+// replaces the SA token in VolumeContext with "[REDACTED]" without mutating
+// the original request or dropping unrelated context keys.
+func TestRedactSATokens_NodePublishVolumeRequest(t *testing.T) {
+	req := &csi.NodePublishVolumeRequest{
+		VolumeId: "vol-123",
+		VolumeContext: map[string]string{
+			metadata.SATokenVolumeContextKey:         testToken,
+			"csi.storage.k8s.io/pod.name":            "my-pod",
+			"csi.storage.k8s.io/serviceAccount.name": "my-sa",
+		},
+	}
+
+	result := redactSATokens(req)
+
+	sanitized, ok := result.(*csi.NodePublishVolumeRequest)
+	if !ok {
+		t.Fatalf("redactSATokens returned %T, want *csi.NodePublishVolumeRequest", result)
+	}
+	if sanitized.VolumeContext[metadata.SATokenVolumeContextKey] != "[REDACTED]" {
+		t.Errorf("got %q, want [REDACTED]", sanitized.VolumeContext[metadata.SATokenVolumeContextKey])
+	}
+	if sanitized.VolumeContext["csi.storage.k8s.io/pod.name"] != "my-pod" {
+		t.Error("non-sensitive volume context keys were dropped")
+	}
+	if sanitized.VolumeId != "vol-123" {
+		t.Error("VolumeId was altered")
+	}
+	// Original must not be mutated.
+	if req.VolumeContext[metadata.SATokenVolumeContextKey] != testToken {
+		t.Error("redactSATokens mutated the original request")
+	}
+}
+
+// TestRedactSATokens_noToken_returnsOriginal verifies that redactSATokens
+// returns the original pointer unchanged when no SA token key is present,
+// avoiding unnecessary allocations on the common path.
+func TestRedactSATokens_noToken_returnsOriginal(t *testing.T) {
+	req := &csi.NodePublishVolumeRequest{
+		VolumeId:      "vol-456",
+		VolumeContext: map[string]string{"csi.storage.k8s.io/pod.name": "other-pod"},
+	}
+	if redactSATokens(req) != req {
+		t.Error("expected original pointer when no SA token key present")
+	}
+}
+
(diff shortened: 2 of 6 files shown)


driver/server_test.go
```diff
@@ -0,0 +1,278 @@
...
+	return c.buf.String()
+}
+
+type capturingSink struct {
+	capture *logCapture
+	kvs     []any
+}
+
+func newCapturingLogger() (logr.Logger, *logCapture) {
+	cap := &logCapture{}
```
wallrj-cyberark: Consider replacing the custom `capturingSink` / `logCapture` implementation (~80 lines above) with [`ktesting`](https://pkg.go.dev/k8s.io/klog/v2/ktesting) from `k8s.io/klog/v2`, which csi-lib already depends on (v2.140.0).

For tests that need to assert on log output, use `BufferLogs(true)`:

```go
logger := ktesting.NewLogger(t, ktesting.NewConfig(ktesting.BufferLogs(true)))
// ... run code that logs ...
testingLogger, _ := logger.GetSink().(ktesting.Underlier)
output := testingLogger.GetBuffer().String()
// assert output does not contain token
```

For tests that just need a logger without inspecting output:

```go
logger := ktesting.NewLogger(t, ktesting.DefaultConfig)
```

This pattern is already established across the cert-manager org:

- **cert-manager/cert-manager** — [`route53_test.go`](https://github.com/cert-manager/cert-manager/blob/master/pkg/issuer/acme/dns/route53/route53_test.go#L147) uses `BufferLogs(true)` + `GetBuffer()` for log output assertions (closest analogue to this PR's use case)
- **cert-manager/csi-driver-spiffe** — [`file_test.go`](https://github.com/cert-manager/csi-driver-spiffe/blob/main/internal/csi/rootca/file_test.go) uses `ktesting.DefaultConfig` for simple logger injection
- **cert-manager/trust-manager** — [`validation_test.go`](https://github.com/cert-manager/trust-manager/blob/main/pkg/webhook/validation_test.go) uses `ktesting.NewTestContext(t)`
- **cert-manager/approver-policy** — controller and webhook tests use `ktesting.DefaultConfig`

`ktesting` handles `logr.Marshaler` and `fmt.Stringer` rendering the same way production backends do, so the token-leak assertions remain valid.

driver/server.go
```diff
@@ -102,7 +104,8 @@ func parseEndpoint(ep string) (string, string, error) {
 
 func loggingInterceptor(log logr.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
-		log := log.WithValues("rpc_method", info.FullMethod, "request", protosanitizer.StripSecrets(req))
+		sanitized := protosanitizer.StripSecrets(redactSATokens(req))
```
wallrj-cyberark, line 107: Re Copilot's performance concern: the eager `protosanitizer.StripSecrets(req)` call was already present before this PR — `redactSATokens` adds only a type assertion and map lookup for non-`NodePublishVolumeRequest` RPCs (negligible). The `protopb.Clone` cost only applies to requests that actually carry a token, which is exactly when redaction is needed.

Deferring sanitisation until a log line is actually emitted would require restructuring how `WithValues` captures the request (logr evaluates eagerly), and is a pre-existing concern outside the scope of this security fix.

driver/server.go
```diff
@@ -113,3 +116,19 @@ func loggingInterceptor(log logr.Logger) grpc.UnaryServerInterceptor {
 		return resp, err
 	}
 }
+
+// redactSATokens returns a copy of req with the ServiceAccount bearer token
+// removed from the VolumeContext of a NodePublishVolumeRequest. All other
```
wallrj-cyberark: Agree with Copilot — the comment says "removed" but the implementation replaces with `[REDACTED]`. Suggest:

```go
// redactSATokens returns a copy of req with the ServiceAccount bearer token
// redacted in the VolumeContext of a NodePublishVolumeRequest. All other
// request types are returned as-is. The original request is never mutated.
```

metadata/metadata.go
```diff
@@ -53,3 +59,21 @@ func FromNodePublishVolumeRequest(request *csi.NodePublishVolumeRequest) Metadat
 		VolumeMountGroup: request.GetVolumeCapability().GetMount().GetVolumeMountGroup(),
 	}
 }
+
+// MarshalLog implements logr.Marshaler so that structured loggers redact the
+// SA bearer token from VolumeContext before emitting a log entry. Without
+// this, log.Info("...", "metadata", meta) would print the live
+// ServiceAccount token injected by kubelet.
+func (m Metadata) MarshalLog() any {
```
wallrj-cyberark, line 68: ~Retracted~ — my original suggestion about using `type noMethod Metadata` to prevent infinite recursion was incorrect.

No logr backend (`funcr`, `zapr`, klog text) recurses on the result of `MarshalLog()`. They all call it once, replace the value, and format the result via reflection. Since none of `Metadata`'s fields are themselves `Metadata`, there is no recursive `MarshalLog()` call path.

The [`type or ObjectRef` pattern in klog](https://github.com/kubernetes/klog/pull/266) solves a different problem: `ObjectRef` has a `String()` method, and without the alias, structured loggers would format it as a flat string instead of JSON fields. `Metadata` has no `String()` method, so this does not apply here.

Apologies for the noise.

wallrj-cyberark (commented):
## Review Summary

Good security fix addressing a real CWE-532 issue — SA bearer tokens appearing in structured log output at default verbosity. The three-sink approach (gRPC interceptor redaction, `logr.Marshaler` on `Metadata`, constant extraction) is sound, and the tests verify both the security property and that the handler still receives the original unredacted token.

### Copilot's comments

- **Performance (line 109)**: The eager `protosanitizer.StripSecrets(req)` call pre-dates this PR — `redactSATokens` adds only a type assertion and map lookup for non-`NodePublishVolumeRequest` RPCs, which is negligible. Not a regression. See inline comment.
- **Doc wording (line 122)**: Valid — "removed" should say "redacted". See inline comment.

### Main suggestion

Replace the custom `capturingSink` with `ktesting` from `k8s.io/klog/v2` — already a dependency of csi-lib, and widely used across the cert-manager org for exactly this purpose. See inline comment with prior art links.

metadata/metadata.go
```diff
@@ -53,3 +59,21 @@ func FromNodePublishVolumeRequest(request *csi.NodePublishVolumeRequest) Metadat
...
+	}
+	redacted := m
+	vc := make(map[string]string, len(m.VolumeContext))
+	for k, v := range m.VolumeContext {
+		vc[k] = v
+	}
+	vc[SATokenVolumeContextKey] = "[REDACTED]"
+	redacted.VolumeContext = vc
+	return redacted
+}
```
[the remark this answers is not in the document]
wallrj-cyberark: This is not correct — no logr backend (`funcr`, `zapr`, klog text) recurses on the result of `MarshalLog()`. They all call it once, replace the value, and format the result via reflection. Since none of `Metadata`'s fields are themselves `Metadata`, there is no recursive call path.

See [my earlier comment](https://github.com/cert-manager/csi-lib/pull/246#discussion_r3465754579) where I investigated this and retracted the same suggestion. The `type noMethod` pattern in klog's [`ObjectRef`](https://github.com/kubernetes/klog/pull/266) solves a different problem: preventing `ObjectRef.String()` from flattening structured fields into a single string in structured backends.

driver/server_test.go
```diff
@@ -0,0 +1,210 @@
...
+// The token must never appear in the string that any logr backend writes to disk.
+
+// TestLoggingInterceptor_errorPath_noTokenLeaked covers Sink #3 (error level,
+// default verbosity). This fires on any NodePublishVolume handler failure,
+// which includes routine conditions such as readOnly not set or issuer timeout.
+func TestLoggingInterceptor_errorPath_noTokenLeaked(t *testing.T) {
+	logger := ktesting.NewLogger(t, ktesting.NewConfig(ktesting.BufferLogs(true)))
+	underlier, _ := logger.GetSink().(ktesting.Underlier)
+	interceptor := loggingInterceptor(logger)
+
```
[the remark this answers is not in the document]
wallrj-cyberark: Valid defensive practice in general, but the assertion can never fail here — `ktesting.NewLogger` always returns a logger whose sink implements `ktesting.Underlier`. A panic with a stack trace would be perfectly diagnostic in test code.

Low priority — up to the author whether to add the check.

driver/server_test.go
```diff
@@ -0,0 +1,210 @@
...
+func requireNoTokenInOutput(t *testing.T, output string) {
+	t.Helper()
+	for _, needle := range []string{"secret", "eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9"} {
+		if strings.Contains(output, needle) {
+			t.Errorf("SA token leaked into log output (found %q):\n%s", needle, output)
+		}
+	}
+}
+
+func bufferOutput(t *testing.T, logger ktesting.Underlier) string {
```
wallrj-cyberark: This helper does not add value over calling `underlier.GetBuffer().String()` directly — it is the same length, and the `t.Helper()` call has no effect because `bufferOutput` never calls `t.Fatal`/`t.Error`. Suggest removing it and inlining the call at each site.

driver/server_test.go
```diff
@@ -0,0 +1,210 @@
...
+	t.Helper()
+	return logger.GetBuffer().String()
+}
+
+// ---- redactSATokens unit tests --------------------------------------------
```
wallrj-cyberark: The `// ---- ... ---` section dividers are noise — they do not help future maintainers understand *why* these tests exist.

Each test function (including `TestRedactSATokens_NodePublishVolumeRequest`, `TestRedactSATokens_noToken_returnsOriginal`, and `TestRedactSATokens_nonNodePublishRequest_returnsOriginal`) should have a short doc comment explaining what security property or invariant it guards. The `TestLoggingInterceptor_*` functions already do this well — apply the same pattern to the `redactSATokens` unit tests.

metadata/metadata_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+const testToken = `{"":{"token":"eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.secret","expirationTimestamp":"2099-01-01T00:00:00Z"}}`
+
+// ---- MarshalLog correctness -----------------------------------------------
```
wallrj-cyberark: Same as above — replace the `// ---- MarshalLog correctness ---` divider with doc comments on the individual test functions. `TestMetadata_MarshalLog_redactsSAToken` and `TestMetadata_MarshalLog_noToken_passthrough` both lack doc comments explaining *why* they exist. `TestMetadata_MarshalLog_tokenNeverInFormattedOutput` already has a good one.

wallrj-cyberark (commented):
## Follow-up review (force-push 03d3ec0)

The ktesting adoption and doc comment fix both look good — thanks for the quick turnaround, Felix.

### CI failure

`pull-cert-manager-csi-lib-verify` failed because `go mod tidy` produces a diff. This is expected after changing imports (adding `k8s.io/klog/v2/ktesting`, removing `github.com/go-logr/logr` from the test file). Running `go mod tidy` and amending should fix it.

### Copilot's new comments

- **MarshalLog recursion (metadata.go:79)**: Incorrect — no logr backend recurses on the result of `MarshalLog()`. See my reply on that thread.
- **Test type assertions (metadata_test.go:44,66)**: Follow from the incorrect recursion premise — not applicable.
- **Underlier type assertion check (server_test.go:110,139,164)**: Valid defensive practice but very low priority — `ktesting.NewLogger` always returns a sink that implements `Underlier`. Up to you whether to add the check.

### Remaining

Code changes look correct and ready to merge once CI is green. A few minor items inline below.

PR #264, opened 2026-07-24
Increase renewal timeout and add Initializing-state diagnostics 

psasidhar:
Three changes to manager/manager.go:

1. issueRenewalTimeout: 60s → 5min Gives the issuer time to stamp Initializing and ZTS time to sign before the driver times out and abandons the in-flight CertificateRequest.

2. case "Initializing" in handleRequest poll loop Tracks when Initializing is first observed, warns if it persists >1min, and logs elapsed time when the CR is eventually signed or times out.

3. Log in findPendingRequest when an existing CR is skipped as non-resumable Surfaces the one-cycle-delay scenario: if a CR is still Initializing when the retry runs, this log confirms the driver is creating a replacement CR.


psasidhar:
This was opened by mistake. Sorry for the noise. Thanks!

PR #269, opened 2026-08-04
fix: use cmmeta.IssuerReference instead of deprecated cmmeta.ObjectReference

wallrj:
`cmmeta.ObjectReference` is a deprecated alias of `cmmeta.IssuerReference` since cert-manager v1.20 and is removed entirely in cert-manager v1.21.

Renaming now (the alias makes this a no-op with the current cert-manager v1.20.3 dependency) unblocks #258, which bumps cert-manager to v1.21.1 and currently fails to compile because of this. Renovate keeps force-push rebasing #258, so a fix commit pushed to its branch gets clobbered — landing the rename on main instead means the next Renovate rebase of #258 will compile as-is.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SgtCoDFish (approved):
/lgtm
/approve

PR #288, opened 2026-08-24
fix(manager): return an error instead of panicking on a non-PEM issued chain

hamodywe:
Picked up from the open follow-ups in cert-manager/cert-manager#9155:

> Latent nil-pointer panic in `calculateNextIssuanceTime`: the `pem.Decode` result is used without a nil check, so a corrupt stored certificate panics the driver.

## The bug

`pem.Decode` returns a **nil** block when the input contains no PEM data. `calculateNextIssuanceTime` dereferenced it immediately:

```go
block, _ := pem.Decode(chain)
crt, err := x509.ParseCertificate(block.Bytes)   // panics when block == nil
```

`chain` is `req.Status.Certificate` — written by the signer, not by the driver. Any external issuer that marks a `CertificateRequest` `Ready=True` while putting something other than PEM in `status.certificate` (a truncated write, a base64 mix-up, a header-only blob) takes the driver process down instead of failing that one volume. There is no CRD-level schema validation on that field, so nothing upstream of this line rejects it.

The caller already treats the function as fallible — it wraps the result in `calculating next issuance time` — and drivers built on this library call it *before* writing anything, precisely so a failure leaves no partial state on disk (csi-driver-spiffe says so in a comment). A panic bypasses all of that.

## The fix

Return an error when there is no PEM block, matching the wording of the `x509.ParseCertificate` failure right below it. Four lines, no behaviour change for any input that already parsed.

## Verification

- **Revert-checked**: with `manager.go` restored to `main` and the new tests kept, the run ends in `panic: runtime error: invalid memory address or nil pointer dereference`. With the fix, all six cases pass.
- New cases cover nil, empty, non-PEM, a truncated PEM block, and a PEM block whose contents are not DER. The last one is deliberate: it goes through the pre-existing `ParseCertificate` error path, so the test also shows the new branch is not swallowing errors that were already handled.
- The existing case (2/3rds of the lifetime) is unchanged; the table just gained an explicit `chain` field, since it previously ignored its own input.
- `go test ./...` failure set is **identical** to a pristine tree on this machine (`storage` ×4 and `test/integration` ×2 fail on `main` too — Windows checkout, not related to this change). `gofmt` and `go vet` clean.

The same one-line omission exists in csi-driver-spiffe's own copy of this function; filed alongside as cert-manager/csi-driver-spiffe#594.

---

*Written with assistance from Claude (Opus 5); the analysis, tests and verification above were run and checked by me.*

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -981,6 +981,10 @@ func (m *Manager) Stop() {
 // NotBefore).
 func calculateNextIssuanceTime(chain []byte) (time.Time, error) {
 	block, _ := pem.Decode(chain)
+	if block == nil {
+		return time.Time{}, errors.New("parsing issued certificate: no PEM data found in the issued chain")
+	}
+
 	crt, err := x509.ParseCertificate(block.Bytes)
 	if err != nil {
 		return time.Time{}, fmt.Errorf("parsing issued certificate: %w", err)
--- a/manager/manager_test.go
+++ b/manager/manager_test.go
@@ -671,18 +671,40 @@ func Test_calculateNextIssuanceTime(t *testing.T) {
 	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
 
 	tests := map[string]struct {
+		chain   []byte
 		expTime time.Time
 		expErr  bool
 	}{
 		"if no attributes given, return 2/3rd certificate lifetime": {
+			chain:   certPEM,
 			expTime: notBefore.AddDate(0, 0, 2),
 			expErr:  false,
+		},
+		"a nil chain is rejected": {
+			chain:  nil,
+			expErr: true,
+		},
+		"an empty chain is rejected": {
+			chain:  []byte{},
+			expErr: true,
+		},
+		"a chain that is not PEM is rejected": {
+			chain:  []byte("this is not a PEM encoded certificate"),
+			expErr: true,
+		},
+		"a truncated PEM block is rejected": {
+			chain:  certPEM[:len(certPEM)/2],
+			expErr: true,
+		},
+		"a PEM block that is not a certificate is rejected": {
+			chain:  pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: []byte("not DER")}),
+			expErr: true,
 		},
 	}
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-			renewTime, err := calculateNextIssuanceTime(certPEM)
+			renewTime, err := calculateNextIssuanceTime(test.chain)
 			assert.Equal(t, test.expErr, err != nil)
 			assert.Equal(t, test.expTime, renewTime)
 		})


manager/manager.go
```diff
@@ -981,6 +981,10 @@ func (m *Manager) Stop() {
 // NotBefore).
 func calculateNextIssuanceTime(chain []byte) (time.Time, error) {
 	block, _ := pem.Decode(chain)
+	if block == nil {
+		return time.Time{}, errors.New("parsing issued certificate: no PEM data found in the issued chain")
```
wallrj, line 985: Returning an error here leaves the volume stuck on this CertificateRequest until the driver restarts.

The return happens before [`m.deletePendingRequestPrivateKey(req.UID)`](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L642), so the private key stays cached. On the next attempt `issue()` finds the same CR, [`certificateRequestCanBeResumed`](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L528) treats `Issued` as resumable, [the key is found](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L440), and `handleRequest` reads the same bad `status.certificate`. No new CR is ever created. The [renewal loop](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L882) retries with backoff for ever, and kubelet's `NodePublishVolume` retries take the same path.

Before this PR the panic restarted the driver. That emptied the in-memory key map, so the next `issue()` took the ["no private key" branch](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L447), deleted the CR and requested a fresh one. That was the only self-heal path and this change removes it. The same stall already exists for the `x509.ParseCertificate` branch just below, so you are widening it rather than creating it.

Suggested fix, in `handleRequest`: drop the pending key on a parse error so the next attempt starts a new request.

```go
	renewalPoint, err := calculateNextIssuanceTime(req.Status.Certificate)
	if err != nil {
		m.deletePendingRequestPrivateKey(req.UID)
		return fmt.Errorf("calculating next issuance time: %w", err)
	}
```

A manager-level test would catch this. The table test cannot.

<details>
<summary>Evidence: issue() twice with a garbage certificate</summary>

I wrote a test in the `manager` package that starts a `Manager` with `newDefaultTestOptions`, runs `testutil.IssueOneRequest(..., []byte("garbage"), ...)`, registers metadata for one volume, and calls `m.issue()` twice.

- On `main`: `panic: runtime error: invalid memory address or nil pointer dereference`.
- With this PR: both calls return `calculating next issuance time: parsing issued certificate: no PEM data found in the issued chain`. After the second call there is still exactly one CR with the same name, `readPendingRequestPrivateKey(first.UID)` is still true, and `IsVolumeReady` is false.
- Same result on `main` with a `CERTIFICATE` block whose bytes are not DER, which shows the stall predates this PR.
</details>

manager/manager.go
```diff
@@ -981,6 +981,10 @@ func (m *Manager) Stop() {
 // NotBefore).
 func calculateNextIssuanceTime(chain []byte) (time.Time, error) {
 	block, _ := pem.Decode(chain)
```
wallrj, line 983: cert-manager already ships this guard. [`pki.DecodeX509CertificateBytes`](https://github.com/cert-manager/cert-manager/blob/v1.21.1/pkg/util/pki/parse.go#L121) returns an error for nil, empty, non-PEM and truncated input, and adds a 330 kB size bound. It is the dependency in `go.mod` already, and `examples/simple/main.go` already imports it. The body becomes two lines and the six new test cases still pass with it (I checked):

```go
	crt, err := pki.DecodeX509CertificateBytes(chain)
	if err != nil {
		return time.Time{}, fmt.Errorf("parsing issued certificate: %w", err)
	}
```

For the record, it does not check `block.Type` and it uses the first block, the same as `pem.Decode`. One difference: a malformed second block now errors where the current code ignores it.

This would also let [`getCertNotAfterTime`](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/metrics/certificaterequest_collector.go#L195) and [`examples/simple/main.go`](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/examples/simple/main.go#L298) share one decoder instead of three copies with three different messages.

wallrj (commented):
Thank you for picking this up from cert-manager/cert-manager#9155. The guard is right and the revert check is convincing. Two things to change before merge, and three outside the diff that you can take or leave.

Before merge (inline comments):

- The new error path leaves the volume stuck on the same CertificateRequest until the driver restarts. The panic-restart was the only recovery path. See the comment on line 985.
- The new tests pass even with the new branch removed. See the comment on line 708.

Outside the diff:

- [`examples/simple/main.go` line 298](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/examples/simple/main.go#L298) has the same unguarded `pem.Decode`. It is the reference driver that the e2e tests run and that csi-driver-spiffe copied. `pki.DecodeX509CertificateBytes(chain)` is a one-line swap; the import is already there.
- [`metrics/certificaterequest_collector.go` line 146](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/metrics/certificaterequest_collector.go#L146) drops the expiry gauge silently when its own copy of this parser fails. A log line, or emitting 0, would help an operator who hits the issuer bug this PR is about.
- [Lines 993 to 997](https://github.com/cert-manager/csi-lib/blob/3a9d982f80a59f5c8b29710b62e63facf08c523b/manager/manager.go#L993-L997): if an issuer returns `NotAfter` before `NotBefore`, the renewal point lands after expiry. Pre-existing, and cert-manager's `pki.RenewalTime` has the same shape, so a follow-up at most.

[Claude Fable 5.1]

manager/manager_test.go
```diff
@@ -671,18 +671,40 @@ func Test_calculateNextIssuanceTime(t *testing.T) {
...
+			chain:  pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: []byte("not DER")}),
+			expErr: true,
+		},
 	}
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-			renewTime, err := calculateNextIssuanceTime(certPEM)
+			renewTime, err := calculateNextIssuanceTime(test.chain)
 			assert.Equal(t, test.expErr, err != nil)
```
wallrj, line 708: These assertions do not pin the new branch. I replaced the guard with `if block == nil { block = &pem.Block{} }`, which removes the new error and lets `ParseCertificate` fail instead, and all six cases still pass.

Four of the five new cases (nil, empty, non-PEM, truncated) hit the `block == nil` branch. The last one, "a PEM block that is not a certificate", has Type `CERTIFICATE` with non-DER bytes, so it hits the `ParseCertificate` branch. It does not test a wrong block type; a `PRIVATE KEY` block holding certificate DER is accepted.

Suggest: make `expErr` a string and use `assert.ErrorContains`, and rename the last case to say "bytes that are not DER". One error case per branch would be enough.

Nit, since you are in this fixture: lines 657 to 671 are one call to `testcrypto.MustCreateCertWithNotBeforeAfter(t, testcrypto.MustCreatePEMPrivateKey(t), &cmapi.Certificate{Spec: cmapi.CertificateSpec{CommonName: "test.example.com"}}, notBefore, notAfter)` from `github.com/cert-manager/cert-manager/test/unit/crypto`. The repo already uses it in `test/integration/metrics_test.go`, and it drops the `crypto/rand`, `crypto/rsa` and `math/big` imports.

